PR #53, opened 2013-04-27
Update quickstart_for_developers.md

mroderick:
Fixing a typo


janl:
Hah, good catch. :)

It'll go live on the next deply.

Thanks!

PR #56, opened 2013-05-09
bug fix for store.on signatures and addition of findAll

bbn:
<3

### DIFF
--- a/stuff/backbone-hoodie.coffee
+++ b/stuff/backbone-hoodie.coffee
@@ -43,12 +43,12 @@ Backbone.Collection::initialize = ->
   # @fetch()
 
   if type
-    Backbone.hoodie.store.on    "add:#{type}", (id, attributes, options) =>
+    Backbone.hoodie.store.on    "add:#{type}", (attributes) =>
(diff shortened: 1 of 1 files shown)


gr2m:
Thanks :)

bbn:
A pleasure! Question: what is the best way to make use of this backbone-hoodie lib in a hoodie project? Currently I'm just copying the file into my vendor directory.

gr2m:
spontaneously, I'd make an own repo for it, what you think? Wanna start one?

It could have backbone-hoodie.js. We can add the .coffee version. Add a nice readme with some examples?

bbn:
ok cool. I'll do it later today. we remove it from this repo then?

gr2m:
awesome! And yes, I'll remove it then

bbn:
okay, i've got something basic going:
https://github.com/bbn/backbone-hoodie

fixed a few more bugs as well!

gr2m:
this is so cool! I'll soon get back to my own project, which is also based on Backbone, looking forward to use your adapter!

PR #59, opened 2013-05-11
Add CommonJS module declaration, fixes #58

### LINKED ISSUE #58: hoodie.js should expose itself as a commonjs module
suggested by @nnarhinen on IRC.

I've never did that before, but happy to help to put it in.

nnarhinen:


caolan:
Good idea. We might want to look at a 'universal module definition' which supports AMD too (example: https://github.com/umdjs/umd/blob/master/returnExports.js)

nnarhinen:
I didn't add amd support as we would have to wrap everything inside a
closure to inject jquery dependency, which I have understood will be
removed. (jquery supports amd but not commonjs)

gr2m:
Thanks @nnarhinen, looks good!

PR #107, opened 2013-07-26
Closures

gr2m:
_this is work in progress_

I'd like to modularize hoodie.js for simpler testing and better overview. This is just an attempt to do so, I'm very open for alternative suggestions.

Hoodie.extend would look like

```
var extensions = [];
Hoodie.extend = function(extension) {
  extensions.push(extension);
};
```

And the extensions would be loaded like this:

```
function applyExtensions(hoodie) {
  var extension, name;
  for (var i = 0; i < extensions.length; i++) {
    extensions[i](hoodie)
  }
}
```

To test the extensions, I'd make the functions passed to Hoodie.extend accessible globally. And in the build process, wrap the whole hoodie.js into a closure to prevent it from polluting window.

```
function hoodieRequest(hoodie){
  hoodie.request = function() {}
}
Hoodie.extend(hoodieRequest);
```

Then in the specs simply do

```
this.hoodie = new Mocks.Hoodie();
hoodieRequest(this.hoodie);
this.hoodie.request // can now be tested.
```

For an example, check [core/request.js](https://github.com/hoodiehq/hoodie.js/blob/eba4d61694c19bbafe49e9bfc32d12cd185b215c/src/core/request.js) and [request.spec.js](https://github.com/hoodiehq/hoodie.js/blob/eba4d61694c19bbafe49e9bfc32d12cd185b215c/test/specs/core/request.spec.js#L13)


gr2m:
I'd like to make things more modular, e.g. having [the new uuid method](https://github.com/hoodiehq/hoodie.js/issues/89#issuecomment-21595509)  method in its own file and then "require" it from `/src/hoodie.js`. But people are very opinionated on module loaders, so I'd love to have a agnostic solution, that uses one of the common loaders if it's present, otherwise falls back to save the modules on `window`.

Does that makes sense? Any thoughts @svnlto @dilvie

dist/hoodie.js
```diff
@@ -1,4 +1,4 @@
-//  hoodie 0.2.2
+//  hoodie 0.2.3
```
janl, line 1: this is a more than trivial change of code, maybe bump to 0.3.0?

svnlto, line 1: :+1: this ain't a minor change

gr2m, line 1: I didn't bump the version, it was just a `grunt build`. This is all playground mode atm. If we merge it, I'll bump it to 0.3.0

gr2m:
@dilvie @svnlto I've rewritten most of the hoodie core extensions to factories. Could you have a look at it and tell me what you think about it?
https://github.com/hoodiehq/hoodie.js/tree/closures/src/hoodie

the ones that are left are `store.js`, `remote.js`, `local_store.js`, `open.js` and `account_remote.js`. The problem here is that currently  `remote.js` & `local_store.js` inherit from `store.js`, and `account_remote.js` inherits from `remote.js`. 

`store.js` defines the store API that should be the same, no matter if data is persisted locally or remotely. It has some basic validations and also implements `update`, `updateAll`, `findOrAdd` and `removeAll` methods, as these just use the other methods internally. Now, how would I make `remote.js` and `local_storage.js` implement and enhance the API provided by `store.js`, to prevent code duplication?

gr2m:
@svnlto don't worry too much about breaking tests. We'll need to redo many of them anyway, as the current tests do a lot of hacky stuff like stubbing private methods or setting private properties. That's not going to work with the rewrite I'm doing now, so don't worry about it. What is important is that we have all the tests describing all the cases that need to be covered. I think that's good enough

gr2m:
so I made an attempt with the inheritance thing. I really want to get this thing done and out of our way.

I made `hoodieRemoteStore` & `hoodieRemoteBase` methods that are factories for APIs that other modules can build upon. For example, in `local_store.js` it looks like this:

``` js
function hoodieStore (hoodie) {
  // inherit from Hoodies Store API
  var storeBase = hoodieStoreBase(hoodie);
  var store = Object.create(storeBase);

  store.save = function(type, id, properties, options)  {
    storeBase.save.apply(store, arguments);
    /* ... */
  }

  /* ... */

  // expose public API
  hoodie.store = store;
}
```

good?

ericelliott:
No. You're effectively forming an inheritance hierarchy and calling super.

This inheritance stuff is the whole point of stampit. You can use
stampit.compose() to inherit the base functionality.

var remoteStore = stampit.compose(store).methods({/\* remoteStore methods
here */});

This will create a new factory with a prototype that has methods from store
mixed in first. If you pass colliding method names into .compose() or
.methods(), the one you pass in last overrides.

There are lots of similar examples in the stampit docs.

I see you're trying to call super - that's a code smell. Erase it from your
vocabulary and better designs will emerge.

gr2m:
so, if both localStore and remoteStore implement a `find` method, they'd do it separately, without a proper way to build upon a base `find` implementation of store? So you end up with code duplication, which is code smell as well, no?

ericelliott:
For an example of "better design", storebase.save() could do its thing and
then call a hook method which inheriting things can override, or trigger an
event which inheriting objects can listen for and respond to, or it can
kick off an action queue (like resolving a promise) that overriding things
can pass callbacks to...

ericelliott:
Or they can both depend on a module that has the shared "find" code...
There are so many better ways to do it, it's shocking to me that people
still think super is a good idea...

gr2m:
Eric, I appreciate your help. I know you have a strong opinion on how to build complex JS projects and a lot of experience backing it, that's why I asked you for your thoughts in the first place. But even if you find it shocking that other developers have other opinions, please let's keep this constructive. All I want is to live up to the promise that the Hoodie API gives, with a internal architecture that is future proof, maintainable and fast. And for that I want to bring in expertise advisors. 

I try to see the benefits of your suggestion to not use `super` methods at all, but I don't understand how the workarounds you mentioned are better than calling super directly. Please be patient with me, I'm sure I'm still missing some important aspects :)

Let's take a concrete example. [store.js](https://github.com/hoodiehq/hoodie.js/blob/closures/src/hoodie/store.js#L92) implements the `update` method. [local_store.js](https://github.com/hoodiehq/hoodie.js/blob/closures/src/hoodie/local_store.js#L355) uses the exact same logic, the only thing it needs to add is to take the promise that the original `update` method returns and extend it with some extra methods.

``` js
// local_store.js
store.update = function update() {
  return decoratePromise( storeBase.update.apply(store, arguments) );
};
```

To my eyes, this is the most simple thing I can think of, it's clear what's happening. 

Two questions:
1. Can you explain why that's bad practice from your perspective? When and why do we have a disadvantage of such code? Could you give one or two examples?
2. With this concrete example, how would you implement the decoration of promises returned by the original `update` method? How / why is it better?

gr2m:
Eric, If you could find some time, the most helpful thing would be if you could fork the `closures` branch and make a pull request with your suggestions for best practice. Don't worry about tests, just showcase changes that you'd suggest and explain why. That will move the whole thing forwards. What you think?

ericelliott:
Sorry, I didn't mean to sound frustrated with you. I'm frustrated with the lack of educational resources for this stuff. These are well-known (in the OO design and patterns communities, at least) OO design techniques that date back to the early 90's, and more people should be teaching and writing about them. To answer your question, a better solution in this case would be to not override .update() at all. Instead, just make it more configurable:

``` js
var store = stampit.methods({
  updateDecorator: {},
  update: function update() {
    var promise = whatever;
    // do stuff...
    // then decorate the promise:
    stampit.extend({}, promise, this.updateDecorator);
  },
  // ...
});


// elsewhere...
var localStore = stampit.methods({
  updateDecorator: //... put the stuff you want to tack on here
});
```

You can even require a prototype for updateReturnMethods in both
places, so you can just extend it in localStore:

``` js
// something like this
var updateDecorator = require('./updateDecorator');
var store = stampit.methods({
  updateDecorator: updateDecorator,
  //...
});

// elsewhere...
var updateDecorator = require('./updateDecorator');
var localStore = stampit.methods({
  updateDecorator: stampit.extend({}, updateDecorator, {
    // put overrides here...
  }
});
```

gr2m:
I get that. But the problem I have for this particular case is that I don't need the `updateDecorater` in the `store` base API. The ability to decorate returned promises is special to `localStore`. Putting it into `store` would make `remoteStore` inherit from it, and other modules that inherit from `remoteStore`. 

It still feels to me that this approach makes things more complex, just to be "right". My problem here is that I don't see the benefits of these _"OO design techniques that date back to the early 90'"_. Are they more performant? Are they simpler to debug? To test? They certainly are not easier to understand. And I try really hard.

But don't worry about it, I don't want to overstretch your patience.

ericelliott:
The idea is to build extensibility into the store API, so that you're not
forced to tightly couple the .update() method to its super().

Google "call super" for writings and references on this subject.  I
mentioned two real-life challenges in my talk that were both the result of
super() calls.

I could write a whole chapter just on how super() can cause problems,
including the cause of timing bugs, forgetting to call super() if
remoteStore gets inherited from, implicit side effects in super() calls...
The list goes on.

The advantage of building the extensibility into the parent API is that
once you do, you don't have to worry about those potential issues in the
inheriting code.

Also, if you don't like my suggestion of building extensibility into the
prototype, consider again that you could share the shared code as a module
that both .update() methods can call, instead of calling .super().

Gruntfile.js
```diff
@@ -27,21 +27,24 @@ module.exports = function(grunt) {
       },
       dist: {
```
ericelliott, line 28: This concatenation not the ideal way to make your code modular. A better solution is to simply rely on a build tool such as browserify or r.js to compile your files for you. If you use browserify, you get the benefit of having the same module require system on the server and in the browser. Browserify can also output an AMD bundle that people can use with bower if they prefer AMD.

svnlto, line 28: AFAIK, AMD / r.js requires you to ship require.js ( or almond ) included in the output file. can you point out how browserify handles this.

ericelliott, line 28: Well, it does need to provide the capability to import dependencies into the modules that require them. There's no getting around that, because browsers don't have that capability built-in.

Here's what it looks like with browserify budles:

``` js
(function(e){if("function"==typeof bootstrap)bootstrap("stampit",e);else if("object"==typeof exports)module.exports=e();else if("function"==typeof define&&define.amd)define(e);else if("undefined"!=typeof ses){if(!ses.ok())return;ses.makeStampit=e}else"undefined"!=typeof window?window.stampit=e():global.stampit=e()})(function(){var define,ses,bootstrap,module,exports;
return (function(e,t,n){function i(n,s){if(!t[n]){if(!e[n]){var o=typeof require=="function"&&require;if(!s&&o)return o(n,!0);if(r)return r(n,!0);throw new Error("Cannot find module '"+n+"'")}var u=t[n]={exports:{}};e[n][0].call(u.exports,function(t){var r=e[n][1][t];return i(r?r:t)},u,u.exports)}return t[n].exports}var r=typeof require=="function"&&require;for(var s=0;s<n.length;s++)i(n[s]);return i})({1:[function(require,module,exports){

// your package code starts here...
```

As you can see, not a lot of overhead here, and it has the added benefit of automatically managing the dependency graph for you, so you always know that the files are loading in the correct order.

gr2m, line 28: Hoodie core will have an account API, a store API (incl. sync) and some helpers. And you'll always need all of its functionality. I'd still spread them out in separate files as we do at the moment and make sure we can test them separately. But I'm not convinced that we need to include any module loader into the code, I don't think that Hoodie core is complex enough to justify that. 

I think you overestimate the complexity of Hoodie Core, Eric. Just have a look at its current code in the `v0.3-cheesecake` branch. Complexity is not an urgent issue atm. For Hoodie Core, we aim from simplicity, maintainability, and performance, in this order.

ericelliott, line 28: 1. What I'm suggesting is that you _simplify_ what you already have even more, using modules. I'm not talking about adding complexity, I'm talking about removing it.
2. A simple `require()` statement would be a very neat solution to the call `super()` issue we were discussing before.
3. Many of your potential users are likely to be using either AMD or browserify. Using browserify can help you auto-generate a UMD bundle that will work with both.
4. If you chose node-style modules, you can eliminate unnecessary boilerplate code you rely on with the module pattern. That reduces the complexity of every file in the project.
5. Stampit uses Browserify, and it's much less complex than Hoodie. Check out the [source](https://github.com/dilvie/stampit/blob/master/stampit.js). As you can see, it imports half a dozen simple utilities. The small size of Hoodie doesn't mean you should avoid using proper modules.
6. Using browserify doesn't mean that you're users are forced into a module loader solution. You can export a single global variable that works as the namespace for your lib. Browserify can do this for you automatically.

gr2m:
this pull request is obsolete. Check out https://github.com/hoodiehq/hoodie.js/pull/116 for current state

PR #113, opened 2013-08-05
fix events#one by providing the correct callback to unbind

timse:
Found a little mistake in the Events#one method. Currently it calls the Events#unbind with the callback that #one gets called with, not the wrapper that actually binds for the event.
Hope it helps and the pull request is ok like this. I changed the test accordingly.


svnlto:
LGTM, thanks @timse

PR #114, opened 2013-08-06
Enabled lower-case letters in uuids.

chrisekelley:
The comment for Hoodie.prototype.uuid states that uuid's are both numbers and lower-case letters; however, it only creates numeric uuid's.


svnlto:
Good catch. LGTM

/cc @gr2m

gr2m:
ay, looks good, cheers

PR #116, opened 2013-08-11
V0.3 cheesecake

gr2m:
image

The code rewrite is done, the tests are mostly fixed :tada: 
But that doesn't mean that it will work at all. We have to test it extensively in real app context now. So let's give it a run and fix it away if you run into a problem.

Man, I'm so glad we get this out. What a beast.


espy:
Wow, well done! Cheesecake well deserved!

svnlto:
@gr2m excellent work mate. :cake:

svnlto:
Ok there is an issue, the build shouldn't pass if there is breaking tests..

gr2m:
@svnlto we need to wrap the concatenated code into a closure to prevent the global scope from being polluted. Any idea what the simplest way is to do that? jQuery uses extra header.js / footer.js files for that.

svnlto:
@gr2m grunt-concat should be able to do that.

PR #127, opened 2013-09-01
hide "/_api" from public API

### LINKED ISSUE #119: add '/_api' automatically when initializing Hoodie
to make a hoodie instance connected to a remote located on a different domain, we currently need to do

``` js
hoodie = new Hoodie('https//example.com/_api')
```

We should hide the /_api part, that should work already:

``` js
hoodie = new Hoodie('https//example.com')
```

gr2m:
fixes #119


svnlto:
LGTM :ship:

PR #135, opened 2013-09-05
V0.4 peanuts store update or add

### LINKED ISSUE #134: rethink hoodie.store.update(type, id, attrs): how to handle NOT_FOUND?
when I do

``` js
hoodie.store.update('doc', '123', { title: 'new title' })
```

it does 2 things:
1. when doc/123 exists, it sets the new title property.
2. if doc/123 does not exist, it creates doc/123 in store with title set to 'new title'

I have no clue why I implemented it this way, and now I find it rather confusing and error prone. 

I run into this while implementing `hoodie.task.restart(type, id)`. I want to use the `store.update` method internally, but I certainly du not want a new task to be created if the task could not be found.

So I suggest to change the behavior of hoodie.store.update, so that it fails when the object cannot be found. If there is a use case when we actually want that, and the use case is reasonable, we can still add a `updateOrAdd` method. But I really can't think of any at the moment.

Any objections? @janl @caolan @svnlto @espy

gr2m:
resolves #134

if object `doc/123` does not exist, this will now fail with a `not found` error:

``` js
hoodie.store.update('doc', '123', {title: 'funky!'});
```

To get the old behavior, use

``` js
hoodie.store.updateOrAdd('doc', '123', {title: 'funky!'});
```

### DIFF
--- a/src/hoodie/store.js
+++ b/src/hoodie/store.js
@@ -294,6 +294,23 @@ function hoodieStoreApi(hoodie, options) {
   };
 
 
+  // updateOrAdd
+  // -------------
+
+  // same as `.update()`, but in case the object cannot be found,
+  // it gets created
+  //
+  api.updateOrAdd = function updateOrAdd(type, id, objectUpdate, options) {
+    function handleNotFound() {
+      var properties = $.extend(true, {}, objectUpdate, {id: id});
+      return api.add(type, properties, options);
+    }
+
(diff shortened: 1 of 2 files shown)


gr2m:
@svnlto could you [check these](https://github.com/hoodiehq/hoodie.js/pull/135/files)?

pleeeeaaaaassseeeeeee

PR #137, opened 2013-09-05
hoodie.task api

### LINKED ISSUE #125: hoodie.task API
as part of the new plugin architecture, we want to provide a nice API for handling tasks in the frontend as well. At the moment, to create a new task, e.g. to send an email, one has to do:

``` js
hoodie.store.add('$email', emailProperties)
```

This exposes the implementation of how tasks are handled in Hoodie internally, and it shouldn't. Also, tasks don't need the entire store API. For example, the user should not be able to change a task's properties after it was created. Instead, the task has to be cancelled and  a new task with the updated properties has to be started. And it also doesn't make sense that creating a task resolves after the object representing it gets stored locally, instead it should resolve once the task has been finished, or be rejected if there was an error doing so.

To reflect that, and to make clear that tasks are different from data, I'd suggest we give it it's own API:

``` js
// starts the task. Once the worker picked it up, it either gets removed (success) or it gets marked as failed
// Rather than resolving right away when the task object was saved locally, it waits until it gets removed from remote
hoodie.task.start('email', emailProperties)

// stop a running task, adds a `canceledAt` timestamp, then removes it
hoodie.task.cancel('email', 'id123')

// if a task had an error, removes the error so task gets restarted
// if task is pending, cancels it and restarts once the worker canceled it
hoodie.task.restart('email', 'id123')

// and probably
hoodie.task.cancelAll()
hoodie.task.restartAll()

// while we're at it
hoodie.task.start('email', emailProperties, { timeout: 20000 })
```

For a plugin creator, (dream) code would look like

``` js
Hoodie.extend(function(hoodie) {
  hoodie.directMessages = {
    send: hoodie.task('direct-message').start,
    on: hoodie.task('direct-message').on
  };
});
```

Find the latest version at https://github.com/hoodiehq/hoodie.js/tree/v0.4-peanuts-task-api#tasks

Thoughts? @svnlto @caolan @espy @janl

gr2m:
resolves #125 

This is it, baby. This is it.

Time to breathe deeply :)


svnlto:
:ship:

PR #144, opened 2013-10-20
New features by da gee

gr2m:
@svnlto I need that for minutes.io. I thought it would be a good feature in general. What you think?


svnlto:
seems legit.

Issue #148, opened 2013-10-30
Bug: on reconnect, tries to push/pull changes for anonymous user


gr2m:
I'm on it

PR #150, opened 2013-10-30
fixes reconnect bug

gr2m:
see #148


svnlto:
:ship:

gr2m:
after such a fix, shall I create a new tag and publish it?

svnlto:
@gr2m yes, if it's a non breaking change you can bump the 'minor' version on publish.

PR #153, opened 2013-10-31
inline docs and global hoodieEvents

svnlto:


gr2m:
sweet! Do you want to keep the `dev` branch around? Shall it be part of some workflow that I should follow, too?

PR #158, opened 2013-11-03
anonymous signup: triggers 'account:signin', should trigger 'account:signin:anonymous'

gr2m:
I'm on it

### DIFF
--- a/src/hoodie/account.js
+++ b/src/hoodie/account.js
@@ -202,9 +202,17 @@ function hoodieAccount (hoodie) {
   // hasAnonymousAccount
   // ---------------------
 
+  // anonymous accounts get created when data needs to be
+  // synced without the user having an account. That happens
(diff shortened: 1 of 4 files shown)


gr2m:
@svnlto for hot fixes we send pull requests directly against master, then rebase dev on it, right?

gr2m:
I'll fix another related bug on the same branch for the sake of simplicity:
hoodie does not connect to remote on `account:signin:anonymous` event

svnlto:
@gr2m please reopen PR against `dev`

svnlto:
is this a hotfix?

gr2m:
> is this a hotfix?

yes, it's a bug. When you subscribe to `account:signin` event, your callback gets called when you create a task from an anonymous account. It should not do that. And the last commit is to make sure that hoodie still connects to remote in case of an anonymous signup

PR #161, opened 2013-11-05
Rename UUID to generateId

### LINKED ISSUE #140: uuid should be renamed
uuid means something specific to most developers (https://npmjs.org/package/uuid) it makes certain guarantees that this uuid method does not.

Even changing it to uid would solve the problem.

garrensmith:
This fixes #140


gr2m:
Looks good to me – thanks @garrensmith

PR #162, opened 2013-11-05
Run tests on save of spec file

garrensmith:
Update Gruntfile to watch for spec changes and then run tests

### DIFF
--- a/Gruntfile.js
+++ b/Gruntfile.js
@@ -26,8 +26,7 @@ module.exports = function(grunt) {
     },
 
     watch: {
-      //files: ['<%= jshint.files %>'],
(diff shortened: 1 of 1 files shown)


gr2m:
ah, I think the line above (commented out) does the same thing, right?

```
//files: ['<%= jshint.files %>'],
```

I'm not sure why we removed it though ... maybe @svnlto knows?

svnlto:
pretty much, yes.

garrensmith:
Ok I've just pushed a change that uncomments that line and removes the current line.

gr2m:
sweet, cheers!

PR #164, opened 2013-11-07
Memory store

alanshaw:
More to open the discussion than anything else.

I needed an in memory store for the [hoodie blackboard](https://github.com/alanshaw/hoodie-blackboard). It's using [punk](https://npmjs.org/package/hoodie-plugin-punk) and [reactive](https://npmjs.org/package/hoodie-plugin-reactive).
- Punk is just a remote store that anyone can write to
- Reactive uses store `change` events to run and re-run computations. In this case when a line is inserted into the database, the canvas is wiped and all the lines redrawn.

Relying on `change` events from the punk remote is too latent when drawing - I need stuff to appear straight away, and get sync'd in the background. `local_store.js` is tied to localStorage and only visible to the current user, not everyone so no good if you want to draw together.

I'm using all these things here (should be reasonably self explanatory):
https://github.com/alanshaw/hoodie-blackboard/blob/master/www/js/main.js#L7-L10

`memory_store.js` is essentially `local_store.js` without the localStorage bits. I guess it could be done better. If you pass it a `remote` option it'll sync with that remote store, after you call `subscribeToOutsideEvents`. In the blackboard example I'm creating a in memory store called "punk" with syncs with the `hoodie.punk` remote.

`memory_open.js` is similar to `open.js` but it opens a memory store and attaches itself to `hoodie.memory.open`.

So really, if this gets merged I'd be looking to have punk create its own memory store, but I thought it would be beneficial for other plugin authors to do the same or for app developers to create memory stores that don't get sync'd to a remote.

### DIFF
--- a/Gruntfile.js
+++ b/Gruntfile.js
@@ -58,7 +58,10 @@ module.exports = function(grunt) {
           'src/hoodie/account.js',
           'src/hoodie/account_remote.js',
           'src/hoodie/task.js',
-          'src/hoodie/scoped_task.js'
+          'src/hoodie/scoped_task.js',
+
+          'src/hoodie/memory_store.js',
+          'src/hoodie/memory_open.js'
         ],
         dest: 'dist/<%= pkg.name %>.js'
       }
--- a/src/hoodie.js
+++ b/src/hoodie.js
@@ -100,6 +100,9 @@ function Hoodie(baseUrl) {
   // * hoodie.remote
   hoodie.extend( hoodieRemote );
 
+  // * hoodie.memory.open
+  hoodie.extend( hoodieMemoryOpen );
+
 
   //
   // Initializations
--- a/src/hoodie.js
+++ b/src/hoodie.js
@@ -48,7 +48,7 @@ function Hoodie(baseUrl) {
 
   /* global hoodieAccount, hoodieRemote, hoodieConfig, hoodieStore,
             hoodiePromises, hoodieRequest, hoodieConnection, hoodieUUID,
-            hoodieDispose, hoodieOpen, hoodieTask
+            hoodieDispose, hoodieOpen, hoodieTask, hoodieMemoryOpen
   */
 
   // * hoodie.bind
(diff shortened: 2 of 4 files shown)


gr2m:
Hey @alanshaw did you work with someone from the Hoodie team on this? It's pretty cool!
First question: could you make a plugin out of it? I'm not sure about merging it into hoodie core, but having a plugin would let others use it without having to wait on what we decide with this pull request.

I wish I had more time right now to review you blackboard app and the plugins you created, but I need to get stuff ready for a conference and will be traveling until Nov 20th. I hope some of the others can jump in and help out

svnlto:
@alanshaw, that's some really cool work! But I have to agree, this would be great as a plugin. 

Let me know if you have any questions.

alanshaw:
Been looking into releasing it as a plugin, but it's tricky as it uses `hoodieStoreApi` which isn't exposed to plugins.

janl:
@alanshaw Here are the docs for the plugins, including the database/storage API: https://github.com/hoodiehq/hoodie-plugins-api

Sorry if there weren’t obvious to find!

gr2m:
I've added the ticket to the new [v0.5 browserify milestone](https://github.com/hoodiehq/hoodie.js/issues/milestones). If I understand it right, that would expose all the internal hoodie modules to the outside.

Until then, feel free to just hack it in a fork so that you can access `hoodieStoreApi`, so you can go ahead and build a plugin. Then we just make it proper once v0.5 is out

svnlto:
>  If I understand it right, that would expose all the internal hoodie modules to the outside.

No, it won't.

gr2m:
As this is in its own plugin now, can we close this? Please feel free to open an issue for farther discussions, we just try to keep open pull requests to zero

alanshaw:
I'm not happy with it though. I've essentially copied a load of code from hoodie core that'll drift over time as you improve it. Like I said, [hoodieStoreApi](https://github.com/hoodiehq/hoodie.js/blob/master/src/hoodie/store.js#L41) isn't available to plugins and that's why I needed to copy it from the core.

gr2m:
Right you are, I made a follow up issue here: #204

PR #174, opened 2013-11-15
hoodie.account.signUp() -> 409 should include a message of “username already exists”

gr2m:


gr2m:
Args sorry should have gone against `dev` branch. But it's so small and master/dev are pretty equal right now, can I leave it like this? @svnlto

PR #180, opened 2013-11-24
remove artificial timeout before local changes get pushed

gr2m:
The reason why this is in is that I wanted to let people save objects on any keyboard press, just like minutes.io is doing it. But truth is, this does not belong into hoodie core, cause by far not all apps are made this way.

### DIFF
--- a/src/hoodie/account_remote.js
+++ b/src/hoodie/account_remote.js
@@ -106,12 +106,12 @@ function hoodieRemote (hoodie) {
   function subscribeToOutsideEvents() {
 
     hoodie.on('remote:connect', function() {
-      hoodie.on('store:idle', remote.push);
+      hoodie.on('store:dirty', remote.push);
       remote.push();
     });
 
     hoodie.on('remote:disconnect', function() {
-      hoodie.unbind('store:idle', remote.push);
+      hoodie.unbind('store:dirty', remote.push);
     });
 
     hoodie.on('disconnected', remote.disconnect);
--- a/src/hoodie/local_store.js
+++ b/src/hoodie/local_store.js
@@ -25,10 +25,6 @@ function hoodieStore (hoodie) {
   // queue of method calls done during bootstrapping
   var queue = [];
 
-  // 2 seconds timout before triggering the `store:idle` event
-  //
-  var idleTimeout = 2000;
-
 
 
 
@@ -683,8 +679,7 @@ function hoodieStore (hoodie) {
 
 
   //
-  // Marks object as changed (dirty). Triggers a `store:dirty` event immediately and a
-  // `store:idle` event once there is no change within 2 seconds
+  // Marks object as changed (dirty). Triggers a `store:dirty` event
   //
   function markAsChanged(type, id, object, options) {
     var key;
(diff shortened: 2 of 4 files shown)


janl:
Should we make a timeout configurable?

gr2m:
I'd rather not, I'd keep it simple. I feel like I did to many assumptions with the current implementation and would wait until people actually ask for additional features like the timeout

janl:
sounds good.

gr2m:
can we leave this open for a bit, I'd like to discuss it a bit farther.

I realized that it's actually not that simple to implement such a timeout in your app code, because the way Hoodie currently does it, your data is persisted immediately, without sending a `POST` request every single time, instead it waits until the store is idle, for 2s per default.

If I'd move that into my app code, I could wait 2s before I call the `hoodie.store.update` method, but that would mean that if I keep updating an object with intervals < 2s and my browser crashes, my changes would be lost.

This could be solved by removing the timeout for the default, but allowing to pass a `timeout` option to the store methods, like 

``` js
hoodie.store.update('doc', '123', {title: 'new title'}, {timeout: 2000})
```

Let's leave that open and discuss it a bit. I'd appreciate a few more perspectives on it

janl:
I could see a case where people don’t want to specify this on every request to the store, but just would want to do something like `hoodie.store.syncTimeout = 2000` and then all operations on the store have that timeout unless overridden by the `update/delete` call.

gr2m:
I had more thoughts on this. I think hoodie core should support the use case where I want to store everything offline, but optionally setting a timeout if I don't want this changes to be pushed to the backend right away. Plugins cannot hook into hoodie.store to make data persist without triggering the changes to be pushed  to remote.

For example: I want my users to type a longer text. I want to store every letter she types, but I do not want to send a request for every change. Instead I want it to wait for e.g. 2s idle time, then push all local changes together.

So per default, I would make the store methods immediate. And optionally I'd let users pass an attribute. But instead of calling it timeout, I'd call it debounce. Timeout has another meaning in the context of async methods

``` js
hoodie.store.update('doc', '123', {title: 'new title'}, {debounce: true})
```

That should do for now, we could hard code the timeout to 2s as we already do. And if enough people ask for more flexibility, we can still allow to change the default globally or per call with `debounce: 300` etc.

Any thoughts on this?

janl:
I’ve never heard the term “debounce”. The option could be `sync_timeout` to avoid ambiguity.

gr2m:
yeah maybe we find a better name for the option. I know debounce from the jQuery plugin: http://benalman.com/projects/jquery-throttle-debounce-plugin/
Underscore is using it, too: http://underscorejs.org/#debounce

janl:
_plays the backend guy card_

If this is something frontend devs are used to, then it is good idea to keep it. I’m not sure if it is exactly the right scenario though, but I am too tired to decide tonight :)

gr2m:
same here :) Just one idea so I don't forget it: we could call the option `wait`.

But it all feels like we are exposing to much internal stuff here ... hmm gotta sleep over it

gr2m:
closing this in favor #202
Zero open pull-request policy ftw

PR #183, opened 2013-11-26
browserify, the quick & dirty

gr2m:
we do this in 3 steps:
1. migrate the current code, with all it's beauty, to the browserify structure. That already gives us the plus of allow plugin devs to do `require('hoodie/events')` and stuff. Publish as v0.5
2. find a tasty name for 0.6 release
3. Make it propor


gr2m:
One thing that I'd like to keep is to have `Hoodie` available as a global, unless the app is using a module managing thingy, in that case I'd make it available only through the respective AMD method. Is that possible?

gr2m:
and can we somehow expose the `require` method, so plugin devs can do things like `require('hoodie/events')`?

gr2m:
I know get this error

```
$ grunt karma:dev
Running "karma:dev" (karma) task
INFO [karma]: Karma v0.10.5 server started at http://localhost:9876/
INFO [launcher]: Starting browser PhantomJS
INFO [PhantomJS 1.9.2 (Mac OS X)]: Connected on socket _bdix6t4-CsqKsP1Xx0g
PhantomJS 1.9.2 (Mac OS X) ERROR
    Error: Cannot find module './hoodie/account'
    at /var/folders/7k/w2f6gw9n60g_d_7q248lg3h00000gn/T/karma-browerify.js:1
PhantomJS 1.9.2 (Mac OS X): Executed 0 of 0 ERROR (0.156 secs / 0 secs)
Warning: Task "karma:dev" failed. Use --force to continue.

Aborted due to warnings.
```

If you have an idea why that is and could point me to how to fix that, I can keep hacking on it

svnlto:
@gr2m no idea

PR #190, opened 2013-11-30
adding /lib to repo again

gr2m:


gr2m:
@svnlto we need to have this discussion, rationally.

I understand it's your preference to not have bower dependencies checked into the repository.
On the other side, I see that causing several problems. 

### Setup for new devs

```
$ git clone git@github.com:hoodiehq/hoodie.js.git
$ cd hoodie
$ npm install
$ grunt test
Fatal error: module "extend" not found from "/Users/gregor/JavaScripts/hood.ie/hoodie.js/src/hoodie/account.js"
# finding out what the problem is ... 
$ npm install -g bower
$ bower install
$ npm test
```

vs

```
$ git clone git@github.com:hoodiehq/hoodie.js.git
$ cd hoodie
$ npm install
$ grunt test
```

### Failing tests

I had several times the problem that tests stopped running, caused by changes in the dependencies in /lib. It was not always obvious and I lost a lot of time on this. I do not want potential contributors to ever have such frustration with hoodie, especially not if there is a way to prevent that.

### GH pages doesn't work any more

http://hoodiehq.github.io/hoodie.js/. It doesn't load the prism dependencies, and I also can't load Hoodie anymore. I used that a lot introduce Hoodie to frontend devs. And github pages is just an example, I just want the index.html page to work out of the box, I put it there for a reason.

---

Maybe I'm missing something, but from my perspective, putting `/lib` into `.gitignore` caused lots of trouble on the one and has little benefit on the other side. Please prove me wrong, or remove it from `.gitignore` again.

Alternatively we could get rid of bower dependencies altogether and use npm exclusively for everything.

svnlto:
> Alternatively we could get rid of bower dependencies altogether and use npm exclusively for everything.

Would be great, unfortunately prism isn't on npm ( Ideally I'd like to get rid of it as it really doesn't add much to the repo other than a bit of syntax highlighting ).

gr2m:
closing in favor of #203

PR #201, opened 2013-12-05
on signIn, $config/hoodie . createdBy gets set to wrong value

gr2m:
The reason is, that hoodie.store prevents the user from overwriting the `createdAt` value, which makes sense. On `signout` event, a new `$config/hoodie` doc gets created and the `createdBy` gets set. Right after on `signin` event, the `createdBy` property cannot be overwritten anymore.

I think the entire way hoodie.config is implemented does not make much sense for several reasons and won't work if we move pouch. But I'll try to fix that specific issue only for now


gr2m:
obsolete via #199

PR #197, opened 2013-12-06
Password Reset: clean up when it failed

gr2m:
when a password reset failed, it should be removed with another PUT request from the client, after the error has been triggered.


gr2m:
I think we should move some of this logic to the backend at some point, there needs to be cleanup routine anyway.
But this works for now, and it's better than nothing.

PR #207, opened 2013-12-11
changing username does not work

gr2m:
The reason is that `salt` gets removed when `$newUsername` gets set


gr2m:
There is another issue with changing usernames. After the existing user object has been removed, we try to sign in to the new account right away:
https://github.com/hoodiehq/hoodie.js/blob/master/src/hoodie/account.js#L1011

`delayedSignIn` works fine if the user account exists, but if the server doesn't create the new user doc immediately, the server returns a 401 and `delayedSignIn` will fail.

Best would be to try to sign in to old account, until it returns the 401, and then sign in to the new account.

gr2m:
manually merged into master.

PR #199, opened 2014-01-15
replace hoodie.account.ownerHash with hoodie.id()

gr2m:
see https://github.com/hoodiehq/hoodie.js/blob/master/src/hoodie/account.js#L1112

And I like the API, every user, no matter if signed up or not, has a **Hoodie ID**, that persists, and that can be used by apps or plugin devs for whatever.
- [ ] hoodiehq/hoodie-plugins-manager#12
- [ ] hoodiehq/hoodie-plugin-users#12


gr2m:
this needs full stack testing, do not merge yet.

gr2m:
This is ready to be merged. Not that https://github.com/hoodiehq/hoodie-plugins-manager/pull/12 & https://github.com/hoodiehq/hoodie-plugin-users/pull/12 have to be merged at the same time, as renaming `ownerHash` to `hoodieId` affects the backend as well.

Also:
1. The now called hoodieId has up till now been locally stored in `$config/hoodie`
   as `_account.ownerHash`, but now as `_hoodieId`.
2. All /_users docs have the property `ownerHash`, it has now be renamed to `hoodieId`

That means that in order to update hoodie.js on current applications, both the CouchDB
and localStorage should be migrated. We do not have a feature build in for that at
this point, so it's in everyone's responsibility to do so. But we should consider to
add such a feature for the future.

janl:
> That means that in order to update hoodie.js on current applications, both the CouchDB
> and localStorage should be migrated. We do not have a feature build in for that at
> this point, so it's in everyone's responsibility to do so. But we should consider to
> add such a feature for the future.

@gr2m can you make a ticket for that?

gr2m:
@janl I got them here: hoodiehq/hoodie-plugins-manager#12 & hoodiehq/hoodie-plugin-users#12

I've also added them as todo's in this pull request's description

janl:
@gr2m I meant the migration bit.

gr2m:
closing in favor of #229

PR #232, opened 2014-02-01
Bootstrapping error handling is missing

gr2m:
After signing in, hoodie starts bootstrapping data from remote and enqueues all `hoodie.store` calls until bootstrapping is done.

Problem is, that if my bootstrapping request fails (e.g. I sign out right after signing in), `hoodie.store` remains in in bootstrapping mode and won't ever dequeue the `hoodie.store` methods, they will stay pending forever.


gr2m:
that's what I'd expect to happen

``` js
hoodie.account.signIn(username, password).then( function() {
  hoodie.store.findAll().fail( function(error) {
    // should be called with HoodieError: bootstrapping aborted
  });
  hoodie.account.signOut()
});
```

gr2m:
@svnlto can we ship that fix?

PR #236, opened 2014-02-04
remove hoodie.generateId() from public API

gr2m:
as we moved generateId into `/utils` folder, I'd remove it from the public API and make it later on accessible to plugin developers using the Hoodie.extend method


gr2m:
@svnlto :ship: or :facepunch:?

Issue #118, opened 2013-08-12
move request pooling into hoodie.js core


gr2m:
I made `withSingleRequest` & `withPreviousRequestsAborted` for request pooling in the  `hoodie.account` pooling.

I think `withSingleRequest` should be the default behavior of GETs / DELETEs and `withPreviousRequestsAborted` the default behavior for PUTs / POSTs in  `hoodie.request`.

That means if I call `hoodie.request('GET', '/_session')` several times before the first request finishes, the subsequent calls don't send another request, but simply get the promise of the first.

svnlto:
+1

gr2m:
closing in favor of https://github.com/hoodiehq/hoodie-client/issues/81

PR #238, opened 2014-02-11
Ensure withSingleRequest is being used for all GET requests

chrisdwheatley:
As per #118 I checked all the current uses of `withSingleRequest` and `withPreviousRequestsAborted` to ensure they we're tied up to the relevant request types. There was only one which wasn't, for `passwordResetStatus`.

### DIFF
--- a/README.md
+++ b/README.md
@@ -165,7 +165,7 @@ You can also subscribe to the following task events
   });
 
   magic.fail(function(eror) {
-    alert('Sory, but something went wrong: ' + error.reason);
+    alert('Sorry, but something went wrong: ' + error.reason);
   });
 
 
--- a/src/hoodie/account.js
+++ b/src/hoodie/account.js
@@ -472,7 +472,7 @@ function hoodieAccount(hoodie) {
       }
     };
 
-    return withPreviousRequestsAborted('passwordResetStatus', function() {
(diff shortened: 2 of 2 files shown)


gr2m:
Thanks for another pull request, Chris!

As a rule of thumb, you want `GET` requests to be `withSingleRequest`, and the other request types to be `withPreviousRequestsAborted`. But in that case I think™ there was a reason why we aborted an eventual previous request, I can't remember what it was though, shame we didn't comment it. Now I can't think of it anymore.

Your change looks good to me at first sight, did you test it in a real app environment?

svnlto:
LGTM, :ship:

chrisdwheatley:
I tested using my-first-hoodie but referencing the built `hoodie.js` I'd amended. Would be interesting to know how you guys feature/regression test within a hoodie app.

svnlto:
Currently we run integration tests like so https://github.com/hoodiehq/hoodie-integration-test unfortunately the test suite is incomplete to say the least. It'd be great if we could come up with a number of scenarios and then run through those on every build. But without derailing this thread, one should probably open another issue ( @swirlycheetah ;)).

/cc @janl @gr2m

gr2m:
I checked it against my own app, looks good. And if we run into trouble, we at least will know why and should make sure to comment it :) Cheers

PR #240, opened 2014-02-16
[refact] utils cleanup

gr2m:
- move utils/request to hoodie/request.
- Remove promise methods from public API (`hoodie.defer()`, `hoodie.reject()`, ...)
- pass lib & util to Hoodie.extend


gr2m:
@svnlto This is ready from my side. This ships the new Hoodie.extend API that we need for hoodie.admin

gr2m:
@svnlto You not happy with the Hoodie.extend API? `Hoodie.extend(function(hoodie, lib, util) {`?

svnlto:
> You not happy with the Hoodie.extend API? Hoodie.extend(function(hoodie, lib, util) {

For now we can go with it (it'll be fine for 1.0) but I think this needs some more work.

steve21124:
Removing hoodie.defer cause problem with other plugin api that use hoodie.defer. like https://github.com/inator/hoodie-plugin-social/blob/master/hoodie.social.js. So how to fix it without hoodie.defer? or need to install additional plugins?

gr2m:
Yeah we removed it from the public API, but it's still accessible to plugin developers, like so:

``` js
Hoodie.extend(function(hoodie, lib, utils) {
  utils.promise.defer // same as hoodie.defer before
});
```

PR #247, opened 2014-03-03
removed dispose function and references to dispose throughout

### LINKED ISSUE #227: remove hoodie.dispose()
... until we actually have a need for it

lewiscowper:
Should fix #227. If you just want stuff commented out, then that's fair enough, I can change that, but here it is with dispose() removed.


gr2m:
Looks good

chrisdwheatley:
:+1:

svnlto:
Looks good tho, just checking with the build to see why it's error'ing again.. Probably down to saucelabs..

Issue #226, opened 2014-01-26
fix comments for account.has(In)ValidSession


gr2m:

lewiscowper:
I think this can be closed based on #248 being merged now.

PR #248, opened 2014-03-03
Comment fix to (in)valid session

lewiscowper:
I'm not sure if this is any better or worse as regards #226.

It looks right, but I've not done a huge amount of digging into it.

My main question would be whether the idea behind:

if (!account.hasAccount())

Is geared toward whether a user is signed into their account, or whether the user has an account at all? If a user has an account but is not signed in, data can be synchronized behind the scenes for them, but I wouldn't want to be synchronizing anything to a user that wasn't signed in, for fear of shared computers and so on.

### DIFF
--- a/src/hoodie/account.js
+++ b/src/hoodie/account.js
@@ -90,8 +90,7 @@ function hoodieAccount(hoodie) {
   // hasValidSession
   // -----------------
 
-  // returns true if the user is currently signed but has no valid session,
-  // meaning that the data cannot be synchronized.
+  // returns true if the user is signed in, and has a valid cookie.
   //
   account.hasValidSession = function() {
     if (!account.hasAccount()) {
@@ -105,8 +104,7 @@ function hoodieAccount(hoodie) {
   // hasInvalidSession
   // -----------------
 
-  // returns true if the user is currently signed but has no valid session,
-  // meaning that the data cannot be synchronized.
+  // returns true if the user is signed in, but does not have a valid cookie 
   //
   account.hasInvalidSession = function() {
     if (!account.hasAccount()) {


gr2m:
@tenmilestereo your comments are not quite correct, but thanks for it anyway, it just confirms that these methods are rather confusing.

The comments should rather read:

``` js
// hasValidSession
// -----------------

// returns true if the user is signed in and session to backend (cookie) is valid.
//
account.hasValidSession = function() {
  if (!account.hasAccount()) {
    return false;
  }

  return authenticated === true;
};


// hasInvalidSession
// -----------------

// returns true if the user is signed but session to backend (cookie) is _not_ valid.
//
account.hasInvalidSession = function() {
  if (!account.hasAccount()) {
    return false;
  }

  return authenticated === false;
};
```

I think that the naming itself is broken, because naming things is hard™. But the current comments are wrong, your's are not quite right either. If mine above make sense, let's put these in for now, or make them more clear if you have suggestions.

lewiscowper:
What about:

``` javascript
// hasValidSession
// ---------------------

// returns true if the user is signed in, and has a valid cookie.


// hasInvalidSession
// -----------------------

// returns true if the user is signed in, but does not have a valid cookie.
```

gr2m:
much better than what we have at the moment :ship:

lewiscowper:
Should be pushed now. :+1:

gr2m:
Great! Can you squash the commits to one? We don't have a commit guideline yet (in the works™), but we try to tag our commit messages as described here: https://github.com/hoodiehq/discussion/issues/3. In your case, I'd do something like `[minor] fix comments for account.hasValidSession / account. hasInvalidSession`

Please let me know if you need help squashing (here's a good intro: http://gitready.com/advanced/2009/02/10/squashing-commits-with-rebase.html). I just learned that myself only a few weeks ago. Once squashed, you can `git push -f` to your branch, that will keep the commits in `master` branch cleaner

svnlto:
@gr2m @tenmilestereo we do, see #9 https://github.com/hoodiehq/hoodie.js/blob/master/CONTRIBUTING.md#pull-requests

lewiscowper:
Is that okay now? Sorry, I'm not great with git yet, it took me a while to get a rebase sorted.

gr2m:
Perfect, thanks @tenmilestereo

PR #255, opened 2014-03-04
[fix #251] test for identical username on changeUsername

### LINKED ISSUE #251: Catch rename to self
Issue description here: https://github.com/hoodiehq/hoodie-plugin-users/pull/18

lewiscowper:
Hopefully fixes #251.

### DIFF
--- a/src/hoodie/account.js
+++ b/src/hoodie/account.js
@@ -501,8 +501,14 @@ function hoodieAccount(hoodie) {
   // But the current password is needed to login with the new username.
   //
   account.changeUsername = function changeUsername(currentPassword, newUsername) {
-    newUsername = newUsername || '';
-    return changeUsernameAndPassword(currentPassword, newUsername.toLowerCase());
+    if (newUsername !== account.username) {
+      newUsername = newUsername || '';
+      return changeUsernameAndPassword(currentPassword, newUsername.toLowerCase());
+    }
+    return rejectWith({
+      name: 'HoodieConflictError',
+      message: 'Usernames identical'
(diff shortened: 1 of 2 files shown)


svnlto:
w00t!

gr2m:
Looks good!

src/hoodie/account.js
```diff
@@ -501,8 +501,14 @@ function hoodieAccount(hoodie) {
   // But the current password is needed to login with the new username.
   //
   account.changeUsername = function changeUsername(currentPassword, newUsername) {
-    newUsername = newUsername || '';
-    return changeUsernameAndPassword(currentPassword, newUsername.toLowerCase());
+    if (newUsername !== account.username) {
+      newUsername = newUsername || '';
```
janl, line 505: I know it is from the old version, but this looks fishy, why should we set the username to `''`?

gr2m, line 505: so that `newUsername.toLowerCase()` doesn't blow up. It will fail anyway when `newUsername` is falsy

janl, line 505: Got it. Might be worth flagging for a rewrite. The intent of the code is not very clear.

I’d do something like this:

``` js
if (!newUsername) { // catches '' (empty string), false, null, undefined
  return false;
}
var lowerNewUsername = newUsernamme.toLowerCase();
return changeUsernameAndPassword(currentPassword, lowerNewUsername);
```

janl, line 505: Thanks for the explanation! :)

PR #261, opened 2014-03-07
Refactor/config

### LINKED ISSUE #216: rethink hoodie.config
- the current `hoodie.config` API is synchronous, although it's using async hoodie.store API internally. That works today, but will break once hoodie.store is truly async
- `hoodie.config` is only used to persist the current hoodieId, username, last sequence nr (for sync) and a pending passwort reset request. All these are meant to remain local anyway, no sync is intended
- I thought `hoodie.config` would make a nice API that other developers could use, too. But to be honest, this is not based on a real use case and if that comes up, it might not be the right solution for it.

I would suggest we remove the API entirely, have a `config` module that is only accessible internally, and that implements it's own localStorage methods, so it works independently from hoodie.store, and always will be synchronous.

It would also prevent the weird problems that are happening when `_hoodieId` and `createdBy` properties from the `$hoodie/config` object diverge. 

Last but not least, the `$` prefix is now used for task object, which `$config` is not.

svnlto:
fixes #216


gr2m:
What do you think about these
1. I find it confusing to call our wrapper around the native `localStorage` object also `localStorage`. I'd rather call it something bulky like `localStorageWrapper` just to make it clear that it's not the native Object the code is referring to. 
2. I'd move the `patchIfNotPersistant` method into `utils/localstorage` and have the methods patched if localStorage is not persistant. Otherwise we'd need to do it in both `hoodie/store` and `hoodie/config`
3. As the config module won't be exposed any more, I'd move it to `lib/config`

svnlto:
amended PR

gr2m:
so now I'm hitting a wall. We do not want to expose `hoodie.config` anymore, but still the internal config should be initialised once and if other modules want to use it, they should get this instance. 

I've no clue how we would do that. Any idea?

gr2m:
@svnlto Okay I think I got it now. I removed the initialisation of the `config` & `localStorageWrapper` utils from `src/hoodie`. They initialise themselves now, we could easily make them their own little packages. All specs are back to green, too. If you're okay with it, I'd do some integration tests in a real-app environment and then merge it if all looks good.

Let's get this out of the way.

svnlto:
LGTM

svnlto:
oups

gr2m:
merged via 00308a1

PR #263, opened 2014-03-11
check connection request: add ?hoodieId=abc4567

gr2m:
That we one can use the logs to get an idea of how many people are using the app, and also find out when a user has been active. It certainly helps with debugging.

I don't see any downside with this, other thoughts?


janl:
It could be that CouchDB doesn't like the extra parameter. If that's the case we gotta strip it in hoodie-server. Should be ok though.

gr2m:
fixed via 1f8d56f

PR #272, opened 2014-03-23
hoodie.account.signIn(username, password) changes hoodie.id() whet it fails

gr2m:
That's quite a critical bug, I'm looking into it and will start a pull request asap.

Hoodie is signing out before it signs in to clean up all the things before signing in as new user. Problem is that it does all that before it knows if sign in succeeds. I'm sorry, I've no idea how that could slip through, it is obviously wrong.

### DIFF
--- a/src/hoodie/account.js
+++ b/src/hoodie/account.js
@@ -241,64 +241,28 @@ function hoodieAccount(hoodie) {
   // Besides the standard sign in we also check if the account has been confirmed
   // (roles include 'confirmed' role).
   //
-  // When signing in, by default all local data gets cleared beforehand (with a signOut).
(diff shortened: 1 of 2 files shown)


gr2m:
note to self: instead of the `signOutAndSignIn` all logic should go into `handleSignInSuccess`: 
- If there is an existing hoodie.account.username, cleanup all local data and trigger `signout` (without actually doing the request) 
- handle `options.moveData`
- `options.reauthenticated` can be handled within `handleSignInSuccess` and can be removed as extra option

plausible?

gr2m:
@svnlto could you please give this a sanity check? I got it already deployed in my app and it seems to work fine

gr2m:
thanks Sven!

PR #276, opened 2014-04-03
options in account.signIn() should default to empty object

lupomontero:
`account.signIn()` is sometimes invoked without passing it an `options` object (as the 3rd argument), so it should default to an empty object to avoid getting a `ReferenceError`.


gr2m:
good catch, thanks Lupo!

lupomontero:
;-)

PR #278, opened 2014-04-08
when local changes get pushed, `change` events from remote get triggered

gr2m:
To reproduce

```
// make sure that you have an account, if not run hoodie.account.anonymousSignUp()

// subscribe to remote changes
hoodie.remote.on('change', function() { console.log('REMOTE CHANGE', arguments) })

// Add a new object
hoodie.store.add('funk', {})

// wait 2s, then you'll see the `REMOTE CHANGE` in the console
```

This issue will become obsolete once we migrate to Pouch (#8), but I think I can fix this with a rather simple fix.

### DIFF
--- a/src/hoodie/store.js
+++ b/src/hoodie/store.js
@@ -140,7 +140,9 @@ function hoodieStore (hoodie) {
       object = cache(object.type, object.id, object, options);
       defer.resolve(object, isNew).promise();
       event = isNew ? 'add' : 'update';
-      triggerEvents(event, object, options);
+      if (!options.silent) {
+        triggerEvents(event, object, options);
+      }
     } catch (_error) {
       error = _error;
       defer.reject(error.toString());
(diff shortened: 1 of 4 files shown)


gr2m:
Here's my idea:
1. When pushing docs, set all `_rev` ids in a hash
2. When handling changes from remote, ignore all that have `_rev` ids from this hash
3. After handling changes from remote, reset the `_rev` ids hash

That should do the trick, and it's not too complicated to put that in now, although Pouch will resolve this anyway.

janl:
should work. Are we persisting _rev ids somewhere, so we can ignore tasks coming back from the server even if we get a page reload / new page in between?

gr2m:
@svnlto could you please give me a sanity check here? I've tested it with my app and it fixes the problem mentioned above, and doesn't seem to have any side effect. I didn't add any extra tests for that because I'm lazy. And because we'll soon merge Pouch I hope, which make all this obsolete.

svnlto:
:ship:

PR #280, opened 2014-04-09
when connecting, hoodie pushes local changes twices

gr2m:


src/hoodie/remote.js
```diff
@@ -114,7 +114,6 @@ function hoodieRemote (hoodie) {
 
     hoodie.on('remote:connect', function() {
       hoodie.on('store:idle', remote.push);
-      remote.push();
```
gr2m, line 117: the `connect` event gets triggered in `lib/store/remote.connect`. But the method there calls push directly after it bootstrapped, so we can remove it from here, otherwise it gets called twice.

PR #284, opened 2014-04-16
[build] use gulp instead of grunt

robinboehm:
use gulp instead of grunt

TODO
- [ ] Bump
- [ ] Karma Test fails

Need help to fix the karma run. @gr2m 

```
PhantomJS 1.9.7 (Mac OS X) ERROR
  ReferenceError: Can't find variable: sinon
  at /var/folders/69/d7_81n851vd7tr6_48x80mmw0000gn/T/karma-browerify.js:5680
```

Just add sinon to the karma.files?


gr2m:
uff I really don't know out of my head, I think best would be to get rid of my fork and try the workaround as described in https://github.com/xdissent/karma-browserify/pull/14. Could you give that a try?

robinboehm:
@gr2m

Got it right ... you want to stub/mock a special file that is required in a test or required in function of the test?

And all that stuff is part of your workaround?

```
global.stubRequire('src/lib/store/remote', remoteStoreStub); 
global.unstubRequire('src/lib/store/remote');
```

So only for this test you want resolve this with a stub? Not fall all test? correct?

---

 If you like, we can solve it together via skype or sth else .

gr2m:
yeah I think you got it right. Let's take `hoodie.id` as an example. The module is defined as `hoodieId` in `hoodie/id.js`. It requires two things:

``` js
var generateId = require('../utils/generate_id');
var config = require('../utils/config');
```

When I test the module, I want to stub what the two require methods return, but only for this test, in the optimal case, the others should not be affected. 

I really appreciate you are looking into it Robin. I know the tests are quite a mess, would be great to clean it up a bit.

robinboehm:
I think we need to find a solution for one e.g. `hoodie.id` and need to refactor all of that kind.
I'll may give a try tomorrow... atm im building another little hoodie app thats more fun :D

gr2m:
hey robin, what's state on this? @svnlto suggested today to replace karma-browserify with https://github.com/karma-runner/karma-commonjs as that's what the karma team is using themselves, maybe that helps?

robinboehm:
Hey @gr2m 
I've too much workload atm, sorry.
May you just want to check this?
I'm not sure when I'll have some time for this

gr2m:
doesn't hurry, we'll just leave it open for now

gr2m:
we'll stick with grunt, @boennemann is currently updating all our repos with a unified test / build / release setup. Cheers!

robinboehm:
:+1:

PR #295, opened 2014-05-15
add parameter 'options' when calling pushLocalChanges()

merlinran:


gr2m:
good catch, Merlin! Can you add a test? I can help if you have questions

merlinran:
I found it by chance, but can't reproduce it any more.

gr2m:
I got it, cheers!

PR #298, opened 2014-05-20
Fixed typos

pborreli:


src/hoodie/account.js
```diff
@@ -364,12 +364,12 @@ function hoodieAccount(hoodie) {
   // ----------------
 
   // This is kind of a hack. We need to create an object anonymously
-  // that is not exposed to others. The only CouchDB API othering such
+  // that is not exposed to others. The only CouchDB API authorising such
```
pborreli: not sure about this one

src/hoodie/store.js
```diff
@@ -399,7 +399,7 @@ function hoodieStore (hoodie) {
   // -------
 
   // object key index
-  // TODO: make this cachy
+  // TODO: make this catchy
```
lewiscowper: I think this one is supposed to be cachy - as in using the cache, just fyi.

src/utils/generate_id.js
```diff
@@ -3,7 +3,7 @@ var chars, i, radix;
 // uuids consist of numbers and lowercase letters only.
 // We stick to lowercase letters to prevent confusion
 // and to prevent issues with CouchDB, e.g. database
-// names do wonly allow for lowercase letters.
+// names do only allow for lowercase letters.
```
lewiscowper: Probably even be better as "We stick to lowercase letters to prevent confusion and to prevent issues with CouchDB, e.g. database names only allow for lowercase letters, the "do" is redundant.

src/hoodie/account.js
lewiscowper: Looks like it might be "offering such functionality", that seems to make more sense.

lewiscowper:
Other than where I've commented, these changes all LGTM.

:+1:

src/hoodie/store.js
pborreli: ah right, lemme revert it

pborreli:
Done, thanks for your feedback.

lewiscowper:
No worries, thanks for the Pull Requests.

Issue #302, opened 2014-05-24
fix travis / saucelabs setup


gr2m:
```
[13:57:52]  <boennemann>       Failed to start Sauce Connect:
[13:57:52]  <boennemann>       24 May 11:53:57 - Error: failed to connect to tunnel VM.
[13:57:52]  <boennemann>     24 May 11:53:57 - Error: please check logfile for possible errors.
[13:58:09]  <boennemann>     https://travis-ci.org/hoodiehq/hoodie.js/builds/25935308#L2388-L2395
[13:59:42]  <boennemann>     Seems like it's set up manually https://github.com/hoodiehq/hoodie.js/blob/master/.travis.yml#L2
[13:59:58]  <boennemann>     Even though travis offers an addon for this http://docs.travis-ci.com/user/sauce-connect/
```

cc @boennemann / @svnlto

boennemann:
fixed 766f75ade66fdc9b50482a6eb9760c7fc9e587c6

PR #303, opened 2014-05-24
Fix Sauce Labs Configuration

boennemann:
I'm not sure if this fixes it already, but I have to open the PR to run the build on Travis.
I'll ping y'all once it's ready.
#302


boennemann:
Cool, Sauce is running now. The build still fails due to a failing test though, but that's another issue.

@gr2m @svnlto

PR #269, opened 2014-05-26
hoodie.account.username should not be set for anonymous accounts

gr2m:
Today, we have this 

``` js
hoodie.account.anonymousSignUp().then(function() {
  alert(hoodie.account.username) // alerts something like abc4567
})
```

This is confusing. It's technically correct, and we need to store the username towards the backend somewhere, but we should not expose it in our API.

### DIFF
--- a/src/hoodie/account.js
+++ b/src/hoodie/account.js
@@ -125,9 +125,6 @@ function hoodieAccount(hoodie) {
   // to sign in with a 300ms timeout.
   //
   account.signUp = function signUp(username, password) {
-    var options;
-
-
     if (password === undefined) {
       password = '';
     }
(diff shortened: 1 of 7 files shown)


gr2m:
This is really confusing. Not only does `hoodie.account.username` get set to `hoodie.id()`, also `account:signup` does get triggered, and it really shouldn't. The anonymous sign up is an internal hack which primary reason of existence is to allow anonymous users to use tasks without signing up. Technical details of it should not be exposed the way they are right now.

I'll look into it asap.

gr2m:
This is work in progress. Ping me if you want to help. That's what needs to be done
- [x] do not set account.username on anonymousSignUp()
- [x] find another way to return true on account.hasAnonymousAccount()

... I _think_ that should be it

gr2m:
Okay, this went a bit crazy. It was not as simple as it seemed in the first place. This pull request introduces several changes. I'd have separated them out but they were all depending on each other.
- `hoodie.account.username` does not get set to `hoodie.id()`, it remains undefined, until `hoodie.account.signIn / signUp` gets called explicitly. 
- only `hoodie.account.signIn` triggers a `signin` event now. Until now, it was also triggered e.g. by `hoodie.account.signUp`
- `hoodie.account.anonymousSignUp` triggers `signup:anonymous`. Note that it's also called in `hoodie.task.start` for users without accounts
- `hoodie.account.changeUsername / changePassword` now trigger `changeusername` /  `changepassword` events

That's it I think. There were also some bugs that I can't remember anymore. 

The `hoodie.account` API is still rather confusing and needs a cleanup, this pull request is a good start for it and also reduces some confusing by not triggering unexpected events or setting `hoodie.account.username` for users that did not sign up or sign in.

svnlto:
LGTM

PR #308, opened 2014-05-30
travis keeps failing because of saucelabs error

boennemann:
image

Any idea what might causing it? @boennemann @svnlto


boennemann:
@gr2m @svnlto I'm not 100% sure why, but now it works. Please Merge (:

PR #323, opened 2014-06-09
bearer token support

michielbdejong:
This PR does two simple things:
- look for an 'authToken' field in the sign-in response, and
  store this as hoodie.account.authToken
- send an 'Authorization' header with 'Bearer ' + that token in
  each request to the hoodie-server.

This PR goes together with https://github.com/hoodiehq/hoodie-server/pull/288
on the hoodie-server repo, which at the HAPI level, will:
- strip the couchdb cookie and add it into the JSON response of
  signIn requests.
- read the Authorization header to look for a token of the form
  'Bearer ...', and create a cookie header from the '...', to
  send to the couchdb layer.


michielbdejong:
btw, if no authToken is found, then this is backwards compatible and just works with cookies the old way (but then only on same-origin, of course).

svnlto:
LGTM

janl:
would want to review before this gets merged, give me some time :)

gr2m:
I'll review this today / tomorrow

gr2m:
It looks good to me so far, I still have to test it with the backend counterpart.

One thing I'd like to change is to make `authToken` private to the `hoodie.account` module instead of exposing it on `hoodie.account.authToken`. I understand the reason you did that was that you need to be able to access it in the `hoodie.request` module, but it would be good to find a better way. Maybe someone has a suggestion? We can discuss this in a separate issue as well. For example, we could add  `hoodie.request.addCustomHeader(name, value)` & `hoodie.request.removeCustomHeader(name)` methods, that could be called within in `hoodie.account` module.

How gets `hoodie.account.authToken` set on initialization? Right now it looks like it would stop working after a page reload, or am I missing something?

gr2m:
`hoodie.account.authToken` is undefined after page reload, you have to load it from config during the hoodie initialization `src/hoodie.js`

After I sign in, the first request to the user's doc `GET http://127.0.0.1:6013/_api/_users/org.couchdb.user%3Auser%2F<username>` returns 404, the bearer token does not get passed.

gr2m:
@michielbdejong can you please look into the open issues:
1. `hoodie.account.authToken` is undefined after page reload, you have to load it from config during the hoodie initialization `src/hoodie.js`
2. After I sign in, the first request to the user's doc `GET http://127.0.0.1:6013/_api/_users/org.couchdb.user%3Auser%2F<username>` returns 404, the bearer token does not get passed.

michielbdejong:
ok, I will have a look!

michielbdejong:
@gr2m first issue fixed, second not reproduced.

Check out https://hoodie.5apps.com/# (user: testing, pwd: testing) - is there something I can run in the console to reproduce your problem?

michielbdejong:
(https://hoodie.5apps.com/#  doesn't work in Firefox because of Mixed Active Content btw, so use Chrome to test)

gr2m:
I guess solving the first issue resolved the 2nd as well. I'll check it today, thanks Michiel!

gr2m:
Thanks, looks good to me!

I've a question: can I override the `Authorization: Bearer ...` header with a custom authorization header? That's a feature we'll need soon, for password protected shared databases, where my current CouchDB user has no access to, but to which I can success by manually passing a username / password. 

So for example this code

``` js
shareId = 'abcdef';
sharedStore = hoodie.share('shareId', {password: 'secret'});
sharedStore.findAll().done(showSharedObjects);
```

will send a `GET /share%2fabcdef/_all_docs` request with the `Authorization` header set to `abcdef:secret`. A hidden CouchDB user with these credentials will exist.

Do you see any potential problem with that and your solution?

michielbdejong:
> Do you see any potential problem with that and your solution?

you could put the `abcdef:secret` info in a header which you call `X-Share-Authorization` or something similar to avoid the name clash

gr2m:
What I mean is: does the server still accept a GET request with the Authorization header set to `"Basic" + btoa("testing:testing")`? I wanted to test it on https://hoodie.5apps.com/, but your Hoodie Backend seems down

michielbdejong:
@gr2m no, that will probably not work (anymore). But that's not a feature that hoodie uses, right?

gr2m:
> But that's not a feature that hoodie uses, right?

Not Hoodie Core, but plugins. We need a way to authenticate as a different user for shared databases. It must work not signed in users as well as users that are signed in as a different user, without their session to be affected

michielbdejong:
so then those plugins can use a `X-...` header for that, right?

michielbdejong:
"peerinvalid The package browserify does not satisfy its siblings' peerDependencies requirements!" in the Travis build, what does that mean? Doesn't look like it's something caused by the changes this PR makes?

gr2m:
> peerinvalid The package browserify does not satisfy its siblings

This is karma / browserify madness :/ I think this is fixed in master now, could you try to rebase on top of it?

michielbdejong:
The reason the tests are failing is because they're [failing in master as well](https://github.com/hoodiehq/hoodie.js/issues/373#issuecomment-55105104).

michielbdejong:
Will rebase as soon as the `promise_restored` branch lands

gr2m:
> Will rebase as soon as the promise_restored branch lands

Yeah that'd be best I think. Sorry for the trouble, we are somewhat between chairs right now, having this major refactoring in the pipeline. I'll reach out one's it's ready, okay?

michielbdejong:
Actually, why can't it be merged now? After all, it's rebased on current master now. And we have to merge this one before merging the hoodie-server one which relies on this functionality.

This PR has a LGTM [from @gr2m](https://github.com/hoodiehq/hoodie.js/pull/323#issuecomment-47784201) and [from @svnlto](https://github.com/hoodiehq/hoodie.js/pull/323#issuecomment-45584861). AFAICT we were waiting for @janl's comment from 10 June:

> would want to review before this gets merged, give me some time :)

Is that still the current status?

janl:
nope all good from my end

gr2m:
@michielbdejong @janl @svnlto @boennemann Okay, I'll take care of merge. Thanks so much for all the great work, Michiel <3

Issue #296, opened 2014-05-16
inconsistent object type validation


merlinran:
https://github.com/hoodiehq/hoodie.js/blob/master/src/lib/error/object_type.js#L23

``` js
var validTypePattern = /^[a-z$][a-z0-9]+$/;
```

https://github.com/hoodiehq/hoodie.js/blob/master/src/lib/error/object_type.js#L30

``` js
HoodieObjectTypeError.prototype.rules = 'lowercase letters, numbers and dashes allowed only. Must start with a letter';
```

There's no dash in the pattern.

It's worth discussing whether dash, underscore, or camelCase is allowed as type name, and write it down to the documentation.

gr2m:
good point. Back when I set the restrictions for types / IDs, I wanted to make them as restrictive as possible. But meanwhile I think that we should at least allow for `-`, for both. Let's leave that open, maybe get more opinions on it. @janl @caolan Any objections?

janl:
I don’t see an issue with dashes or underscores. I’d not do camel case.

PR #333, opened 2014-06-28
Allow dashes and underscores in type names #296

lupomontero:
A small change, but it does help when having to come up with task names...

Since nobody has raised any objections in #296 I have added both dashes and underscores. I can confirm I can run tasks with names like `do_this` or `do-that-as-well` as they often appear in "docs" and all seems well.

### DIFF
--- a/src/hoodie/store.js
+++ b/src/hoodie/store.js
@@ -777,7 +777,7 @@ function hoodieStore (hoodie) {
 
 
   // a semantic key consists of a valid type & id, separated by a "/"
-  var semanticIdPattern = new RegExp(/^[a-z$][a-z0-9]+\/[a-z0-9]+$/);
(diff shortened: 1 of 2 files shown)


gr2m:
Can just add dashes for now? I'd rather be restrictive, and see how it goes. 

Could you stash your commits and follow [Angular's Commit Style Guide](https://github.com/angular/angular.js/blob/master/CONTRIBUTING.md), as we use them to auto-generate our [Changelog](https://github.com/hoodiehq/hoodie.js/blob/master/CHANGELOG.md).

lupomontero:
@gr2m Thanks for the feedback. I have updated the PR as suggested (hope all is well).

I also realised [this pattern](https://github.com/hoodiehq/hoodie.js/blame/master/src/hoodie/store.js#L780) needed to be updated as well so also included it in the PR.

gr2m:
Thanks Lupo!

PR #338, opened 2014-07-14
hoodie.account.usernameChange does not check for the case of error

gr2m:
If the new username is already taken,  an `$error` property gets set on the user doc, but the promise returned by `hoodie.account.usernameChange` does not get rejected


gr2m:
Please do not merge yet, it's something we needed for internal use, we'll review properly before merging

PR #294, opened 2014-05-15
Prevent multiple connects from resulting in conflicting calls to remote.push

redgeoff:
More details can be found at https://github.com/hoodiehq/hoodie.js/issues/283

PR #355, opened 2014-08-16
remote stores can trigger multiple 'connect' events

gr2m:
`remote.connect` should not trigger connect or bootstrap when it is already connected.


gr2m:
fixed via #294

Issue #343, opened 2014-07-19
Refactor hoodie.js: externalize state for simpler testing


gr2m:
this is in process, see https://github.com/hoodiehq/hoodie.js/commit/b3f53975e468198fc66f831feb3444bc649d11f5 for example.

boennemann:
I'll maintain my todo list here, where I'll check off modules :shipit: 
- [x] hoodie
  - [x] :red_circle: account
  - [x] :red_circle: connection 
  - [x] :red_circle: id 
  - [x] open
  - [x] :red_circle: remote 
  - [x] :red_circle: request 
  - [x] :red_circle: store 
  - [x] :red_circle: task 
- [x] lib
  - [x] error
    - [x] error
    - [x] object id
    - [x] object type
  - [x] store
    - [x] :red_circle: api 
    - [x] :red_circle: remote 
    - [x] :red_circle: scoped
  - [x] task
    - [x] scoped
  - [x] ~~events~~ replaced w/ EventEmitter wrapper
- [x] utils
  - [x] promise
  - [x] :green_heart: events
  - [x] config
  - [x] generate id
  - [x] hoodiefy request error name
  - [x] :green_heart: local storage wrapper

:red_circle:: Tests don't even run before the refactoring.

svnlto:
If you don't mind I steal the above and add test status to the rest.

boennemann:
@svnlto Sure go ahead. You wanna do this right here? I could then focus on refactoring the ones where tests pass (:

svnlto:
> You wanna do this right here? 

will do.

gr2m:
fixed via https://github.com/hoodiehq/hoodie.js/pull/356

Issue #311, opened 2014-06-02
returning standard-compliant promise


gr2m:
Until now, we use jQuery's implementation of promises for now. Reason being that we think that there is (or will be) a big intersection between users jQuery and potential users of Hoodie.

We now decided to go with standard-compliant promises as well. Reasons being
1. Developer happiness (both internal and plugin developers)
2. Integration with other libraries, like Ember, Pouch, etc
3. We can still provide `.done` / `.fail` / `.progress` callbacks that have the same signature as the counterparts of jQuery. And to the end users, these three will be the most important ones, while `.then` is more important do pluging developers

Next steps:
- [x] Find a polyfill library library for browsers that do not have native support for `Promise` (https://github.com/petkaantonov/bluebird)
- [x] Make it a Hoodie dependency
- [x] add `.done` / `.fail`
- [x] add `.progress` callbacks
- [x] replace all internal usage with the new implementation

---

Current work on this lives in the promise branch:
https://github.com/hoodiehq/hoodie.js/tree/promise_restored

calvinmetcalf:
2 options to do:
- use a basic promise library (like my own [lie](https://github.com/calvinmetcalf/lie) or [then/promise](https://github.com/then/promise)) and extend it with your own methods. 
- pick an existing solution that has similar functionality but might have it implemented slightly differently [when](https://github.com/cujojs/when) Is a very fast library that supports progress and I believe it has done style stuff (though might call it finally)

svnlto:
@calvinmetcalf I have started looking into conversion and think that the pouchdb approach would suite us best (congrats on the job well done btw), using bluebird and lie where applicable. 
It'd be really cool if you want to give us a hand actually :smiley:

gr2m:
> It'd be really cool if you want to give us a hand actually

That'd be fantastic. I can't wait to finally replace our self cooked sync solution with PouchDB, and it would be great to have a PouchDB committer helping us out a bit.

calvinmetcalf:
Thanks, I'd be happy to lend a hand, one thing to bear in mind is that
while done and fail are very easy to patch onto browser type promises
progress isn't, PouchDB uses events for this.

svnlto:
Cool. started of by lifting PouchDB's `toPromise` util. https://github.com/hoodiehq/hoodie.js/commits/promise Thoughts?

calvinmetcalf:
I actually have a stand alone version of that one https://github.com/calvinmetcalf/pouchdb-topromise

[...]

PR #356, opened 2014-08-22
The Hoodie Refactor™

boennemann:
This hugenormous PR aims to fix both #311 and #343.

_This is **NOT** yet ready to be merged_ but I'm opening this so we can discuss and track progress.


boennemann:
@svnlto @gr2m 

Together with Gregor I have fixed all the bugs that were caused by the refactor itself, so my-first-hoodie + new hoodie.js at least shows the account bar again.

screen shot 2014-08-22 at 18 55 07

It seems the remaining issues are all promise related. For example `hoodie.account.authenticate()` gets resolved with another promise that fails with undefined, rather than a username (see image). Also it assumes it's logged in even though it's not (cookies & localstorage cleared).

Would be cool if you could look into that, while I'm gone.

How to debug this branch:

``` bash
cd hoodie.js
git checkout promise_restored
npm link
cd -
hoodie new debugApp
cd debugApp
cd node_modules/hoodie-server
npm link hoodie
cd -
hoodie start
```

Oh and if you want to run the integration test suite (more than once) you have to delete testapp before running the tests again. This should (and will) be part of the tests soon. 

``` bash
rm -rf node_modules/grunt-release-hoodie/node_modules/hoodie-integration-test/myapp/
grunt integration-test
```

boennemann:
Oh and @svnlto, 
it would be cool if you could rebase the commits at the beginning of this branch so they have meaningful commit messages :)

gr2m:
I'm looking into this now. Current problem is caused by jQuery, that is still used for requests. Any suggestions on what to use instead?

svnlto:
@gr2m one suggestion would be to use https://www.npmjs.org/package/xhr

janl:
What does pouch use?

svnlto:
https://www.npmjs.org/package/xhr

gr2m:
I fixed the bug causing this:

screen shot 2014-08-22 at 18 55 07

I've added the `done`/`fail`/`always`/`progress` callbacks by extending `Promise.prototype`. I did not yet repleace jQuery's ajax, but wrapped it into a native promise so we can move on, and replace it in parallel later on

gr2m:
Next problem I see is with `config`. `config.get` is not the initialized config instance with internal state, but the one that still expects state as its first parameter. Any idea how to fix that? @boennemann @svnlto

gr2m:
Okay, next up:

``` js
this.hoodie.account.on('signin signup changeusername', handleNewUsername);
```

`handleNewUsername` does not get called with `state.events.emit('signup', username);`, it doesn't split up the events. This works:

``` js
this.hoodie.account.on('signup', handleNewUsername);
```

boennemann:
Seems like an EventEmitter problem. I hope we don't have to hack it in [here](https://github.com/hoodiehq/hoodie.js/blob/promise_restored/src/utils/events.js), but it's just config.

gr2m:
Problem is, EventEmitter does not support binding to multiple events with one call, neither does EventEmitter2.

We want to match jQuery's API. This means
- `module.on(events, handler)`  
  `events`: One or more space-separated event names  
  `handler` gets called any time one of the events gets triggered.
- `module.one(events, handler)`  
  `events`: One or more space-separated event names.  
  `handler` gets only called the first time one of the events gets triggered.
- `module.off(events[, handler])`  
  `events`: One or more space-separated event names.  
  `handler` (optional): The function that is to be no longer executed

Plus add aliases: `bind` -> `on`, `unbind` -> `off`

svnlto:
Here's a good starter, https://github.com/jbrumwell/jqevents/blob/master/index.js

gr2m:
@boennemann I get `stdout maxBuffer exceeded` error for `grunt integration-test` in hoodie.js folder https://gist.github.com/2915cc4ce7b8bbb26828 Any idea how to fix / workaround that?

gr2m:
@boennemann nevermind, I updated grunt-release-hoodie to 2.7.0-rc.4, works now. Also, progress :)

boennemann:
Don't know exactly, but according to the log you haven't deleted the app beforehand:

`rm -rf node_modules/grunt-release-hoodie/node_modules/hoodie-integration-test/myapp/`

gr2m:
With the latest changes, hoodie.js#promise_restored is now passing on Travis:
https://travis-ci.org/hoodiehq/hoodie.js/builds/34582014

boennemann:
Integration tests are running now, awesome :+1:  Thanks @gr2m for pushing the last mile here.

Here is my suggested 10 point plan on how to ship this now:
1. Rebase the entire branch and squash wip commits and change commit messages so they **strictly** adhere to our guidelines (I'll take care of this) 
2. Move current `master` to a `stable` branch so we can still push urgent bug fix releases there.
3. Make this the `master` and release it as `1.0.0-alpha.1` (Actually this version doesn't contain _planed_ breaking changes, but I guess a) it's desirable to have a >1.0 version b) changes are so huge that it's just likely that we'll break something anyways)
4. Test the alpha in apps we have built ourselves
5. Bring unit test coverage to at least 90%
6. Write more integration test scenarios as in https://github.com/gr2m/milestones/issues/27
7. Release `1.0.0-beta.1`
8. Write a ~blogpost where we ask the community to test it with their apps
9. Wait one week
10. Ship `1.0.0`

@janl @svnlto @gr2m What do you think about this?
I could take care of 1-3 immediately if you agree with the plan.

gr2m:
Looks good to me. The `1.0.0` version of hoodie.js isn't the Hoodie 1.0 marketing version, right? Just want to make sure.

I'd move up `Write more integration test scenarios as in gr2m/milestones#27`, I'd do that in parallel to 1 - 5. I'm happy to help writing scenarios, but am currently blocked by https://github.com/hoodiehq/hoodie-integration-test/commit/471a4c791b1d44490d81732567ae2a372df40c5b, would be great if you could look into it.

:shipit:

gr2m:
I could work around the problem at hoodiehq/hoodie-integration-test@471a4c7, see https://github.com/hoodiehq/hoodie-integration-test/commit/b8abb84755d22aacbb0becb260a25c608a9d6545#diff-ead271c307b16870f13ad91f97c698a3R239 and https://github.com/hoodiehq/hoodie-integration-test/commit/b8abb84755d22aacbb0becb260a25c608a9d6545#diff-ead271c307b16870f13ad91f97c698a3R265

gr2m:
@boennemann there is a flaw in `lib/store/remote/*`. We can't use `state.hoodie.remote` in there, these are two different things. For example, `hoodie.open('dbName').find('todo', '123')` currently tries to find `todo/123` in the user's database, instead of in `dbName`. Can you clean that up?

gr2m:
@boennemann besides the foo with `lib/store/remote/*`, we need the progress callback back for promises. Then we're good.

boennemann:
@gr2m `1.0.0` is only referring to hoodie.js, this has nothing to do with the marketing version :)

Also were the recent fixes you pushed present before the refactor happened? Because if they were introduced by the refactor itself they're not a fix, but a refactor, because the commit type should always be relative to the latest release, so the changelog makes sense in die end.

gr2m:
@boennemann I suggest you can squash all the commits regarding the refactor in just one. I just make them separate for easier discussion / collaboration

boennemann:
@gr2m There is no need to squash commits together if they can be separated -> future discussion etc.

But imagine a situation where `1.0.0` of module a is released. You prepare the next version of the module and add a new feature like so: `feat(user): add feature x so you can y`. After pushing this you notice, that the new feature broke something. If you now commit that fix as `fix(account): add back foo for the bar` the changelog for `1.1.0` contains "fix(account): add back foo for the bar" even though it has never been broken in any version before. So are these "real" fixes, or just fixes for bugs that were introduced by the refactor itself?

I'll take care of rebasing the entire branch before making this the master anyways, that's why I'm asking :)

janl:
Good planning & execution here: +1 :)

boennemann:
Okay I'll take care of 1-3 then, by today evening :)

gr2m:
@boennemann @jan shall we  merge the Bearer token pull requests before the rebase? If it passes the current CI tests, it's good to merge

boennemann:
@gr2m Let us take care of this afterwards. I guess that's easier

gr2m:
> we need the progress callback back for promises

If fixed this (for) now. It works™

What's left is the `lib/store/remote/*` foo. Can you make some free time to pair on this? 20min should be enough

gr2m:
@boennemann okay, I think it's all done. `promise_restored` now passes all tests of `hoodie-integration-test`. It's all yours now.

gr2m:
closing in favor of #382

Issue #385, opened 2014-10-23
remove /dist folder & build task


gr2m:
hoodie.js is not meant as an entry point anymore. It depends on and must be loaded from hoodie-server. 

It could, in theory, still be used without the hoodie-server backend, but that's not a primary use case and will definitely cause issues if you still want to use it with a hoodie-server, as they can easily diverge and become incompatible.

svnlto:
:+1:

PR #387, opened 2014-10-26
removed dist folder, and references to it in build tasks

lewiscowper:
Should be enough in here to complete #385


gr2m:
can you also remove `bower.json`? Then squash the commits to just one following our conventions, something like `chore: remove bower & build`

gr2m:
also:

in `Gruntfile.js`
- remove `grunt.registerTask('build...`
- remove `concat` and `uglify` options in `grunt.initConfig`

in `package.json` 
- remove `grunt-contrib-concat` and `grunt-contrib-uglify` from package.json

I think that's it :) Thanks Lewis!

lewiscowper:
I had a bit of git trouble, but that should be it now?

PR #372, opened 2014-11-15
hoodie.task.start should progress with task object

gr2m:
# DO NOT MERGE

I've found an issue with this, I'm on it.

---

it currently progresses with something like

``` js
hoodie.taks.start('test', {}).progress(function(task) {
  // task should be task object with id, type, createdAt etc, but instead is something like
  // {
  //   id: 'org.couchdb.user:user_anonymous/xpyjy50',
  //   rev: '1-72a0f32b32719bca8db387d9e3340184',
  //   ok: true
  // }
});
```

### DIFF
--- a/src/hoodie/task/helpers.js
+++ b/src/hoodie/task/helpers.js
@@ -15,6 +15,10 @@ exports.handleNewTask = function(state, object) {
   var defer = getDefer();
   var taskStore = state.hoodie.store(object.type, object.id);
 
+  taskStore.on('sync', function(object) {
+    defer.notify(object);
+  });
+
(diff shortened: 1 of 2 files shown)


gr2m:
@boennemann This is good to be merged, but to be sure I'd like to test it against the integration tests, incl. this part that is currently commented out: https://github.com/hoodiehq/hoodie-integration-test/blob/master/tests/www/api/task.js#L31-L38

How would you handle such a case? Similar case at https://github.com/hoodiehq/hoodie.js/issues/377

gr2m:
@boennemann ping :)

PR #393, opened 2014-11-29
Handle non-hoodie id's

thomasvs:
If an id is not of the form type/str, log it and keep going.


janl:
@thomasvs any ideas why this would create a merge conflict?

thomasvs:
@janl from a quick browse it looks like master has been significantly rewritten (maybe in the middle of a pouchdb rewrite?) and the file itself in which I made the change is not around anymore.

I started from the 1.0.0 tag as my hoodie-server is currently on 0.9.7 of hoodie.js so that I could have something that works with what you guys currently consider the stable version.

janl:
@thomasvs would it be possible to get this against master?

Issue #390, opened 2014-11-10
hoodie stops loading documents when a document id is not in the expected format, without telling you which doc id


thomasvs:
On loading all docs, if the doc id is not in the format something/somethingelse, hoodie errors out.  It doesn't print on the console which id - it just fails in the line that says

object.type = matches[1], object.id = matches[2]

In addition, it fails even if that wrong doc id is deleted - which imo hoodie should just skip.

gr2m:
I agree, it's a known issue that we just silently ignored, as migrating to PouchDB will resolve it anyway. We started to work on the PouchDB migration if you like to give a hand: https://github.com/hoodiehq/wip-hoodie-store-on-pouchdb

thomasvs:
I think the severity is more than confusing.  In essence, if this happens, you don't get to see all the data that was not synced to local storage after hitting this exception.  This means that to the user the data is effectively inaccessible.

thomasvs:
Here is a fix that works for me: https://github.com/thomasvs/hoodie.js/tree/issue-390

gr2m:
The new hoodie does not require the `type/id` format any more.

PR #395, opened 2014-12-02
Resolve issue #390 against master.

thomasvs:
Not tested.

How can I test this?


janl:
Not an expert on this code base, but would https://github.com/hoodiehq/hoodie.js/blob/master/test/specs/hoodie/id.spec.js make sense? (cc @svnlto)

svnlto:
yup.

thomasvs:
@janl so if I run grunt test on my branch, that test is already part of it?

grunt test passes here, although there are some unrelated-looking ERROR lines on the output.

janl:
@thomasvs yup, it works locally for me. I added a failing assertion and got a fail as expected, so yeah, this seems to work as expected. Which errors are you getting?

gr2m:
The "Possibly unhandled Error" is coming from promises that don't have `.catch` calls. They can be ignored

src/lib/store/remote/helpers.js
```diff
@@ -106,13 +108,21 @@ exports.parseFromRemote = function(state, object) {
   // as in some cases IDs might contain '/', too
   //
   matches = id.match(/([^\/]+)\/(.*)/);
+  if (matches == null) {
+    console.log('hoodie: unhandled id ' + id);
```
gr2m: we can't do `console.log`s, it breaks some browsers. Then it's good to merge, I'll add a test for it in https://github.com/hoodiehq/hoodie-integration-test/. Thanks for your help Thomas!

thomasvs: Ok, removed.  However, it's frustrating that warnings like this can't be logged somehow.  How about adding a hoodie function to log that can be injected for those who care?

gr2m: I absolutely agree. Join the discussion here on the planned `hoodie.log` API https://github.com/hoodiehq/hoodie.js/issues/401

PR #363, opened 2014-12-19
hoodie.open(dbName).disconnect() to return promise

gr2m:
currently returns undefined. I expect it to return a promise just like `.connect()` does


gr2m:
This is good to be merged from my side

PR #376, opened 2014-12-19
handlers of hoodie.store events do not throw errors

gr2m:
this is because we have a try / catch around the hoodie.store's internal triggerEvents method.

### DIFF
--- a/src/utils/events.js
+++ b/src/utils/events.js
@@ -11,10 +11,23 @@ var exports = module.exports = function(hoodie, context, namespace) {
   var emitter = JQEventEmitter.create();
 
   // aliases
-  emitter.trigger = emitter.emit;
   emitter.one = emitter.once;
   emitter.bind = emitter.on;
   emitter.unbind = emitter.off;
+
+  // monkey patch emit with try-catch
+  // because of https://github.com/hoodiehq/hoodie.js/issues/376
+  emitter.trigger = emitter.emit = (function(emit) {
+    return function() {
+      try {
+        emit.apply(emitter, arguments);
+      } catch (error) {
+        setTimeout(function() {
+          throw error;
+        });
+      }
+    };
+  })(emitter.emit);
 
   extend(hoodie, emitter);
 


gr2m:
As we replaced the jQuery Promises with native ones in Hoodie, things got a bit more complicated. Throwing an error is catched by the promise natively and will reject it.

Example

``` js
var promise = new Promise(function(resolve, reject) {
  throw new Error('test')
});
promise.catch(function(error) { console.log(error) });
// Error: test
```

If the promise gets resolved beforehand, errors get silenced, which will make it hard to find and debug these cases.

Example

``` js
var promise = new Promise(function(resolve, reject) {
  resolve("ok");
  throw new Error('test');
});
promise.then(function(value) { console.log(value) });
// ok
```

A solution could be to through the error on next tick. 

``` js
var promise = new Promise(function(resolve, reject) {
  resolve("ok");
  setTimeout(function() { throw new Error('test') }, 0)
});
promise.then(function(value) { console.log(value) });
// ok
// Uncaught Error: test
```

Any thoughts? @svnlto @boennemann 

---

resources
- https://twitter.com/andrewsmatt/status/532569644885102592
- https://twitter.com/andrewsmatt/status/546029674191921152
- https://github.com/matthew-andrews/Promise.prototype.end

gr2m:
@calvinmetcalf have you run into such trouble silenced errors because of promises in PouchDB?

calvinmetcalf:
So promises resolve or reject, never both so you probably want to return after resolving

gr2m:
> So promises resolve or reject, never both so you probably want to return after resolving

Yes that's clear. In Hoodie context, we synchronously trigger events like 'add', and handlers could throw errors. But if these errors get thrown after a promise is resolved or rejected, there is no way to find out about it. It doesn't show up in the console. It leads to bugs that are very hard to debug. In this example below there is no sign about the thrown error in the browser console, it just fails silently.

```
var promise = new Promise(function(resolve, reject) {
  resolve("ok");
  throw new Error('test');
});
promise.then(function(value) { console.log(value) });
// ok
```

Makes sense? Ever run into such a thing with Pouch?

calvinmetcalf:
these sound like things that should be handled with events not errors

janv:
> In Hoodie context, we synchronously trigger events like 'add', and handlers could throw errors. 

Why don't you just trigger the events and resolve the promise afterwards?

jezell:
``` javascript
function $$(fn) {

    return function(resolve, reject) {
        var result, isError;


        fn(function(success) {
            result = success;    
                }, function(error) {
            result = error;
            isError = true;
        });
        return isError ? reject(result) : resolve(result);
    };
}

var promise = new Promise($$(function(resolve, reject) {
  resolve("ok");
  throw new Error('test');
}));
promise.then(function(value) { console.log(value); }).catch(function(err) { console.log(err); });
```

grncdr:
for future readers, I did a bit of digging and the code in question seems to be here now: https://github.com/hoodiehq/hoodie.js/blob/c8ab93dbacab9706a3b43dc37edc12dbc0f202db/src/hoodie/store/localstore.js#L108-L122

> Yes that's clear. In Hoodie context, we synchronously trigger events like 'add', and handlers could throw errors. But if these errors get thrown after a promise is resolved or rejected, there is no way to find out about it.

The code could be rewritten in a more idiomatic promise style like so:

```
return new Promise(function (resolve, reject) {
  object = helpers.cache(state, object.type, object.id, object, options);
  event = isNew ? 'add' : 'update';
  if (!options.silent) {
    helpers.triggerEvents(state, event, object, options);
  }
  resolve(object);
});
```

This is essentially the same as @janv's suggestion ("Why don't you just trigger the events and resolve the promise afterwards?"), but a bit easier to follow.

One thing I'm trying to figure out is why [this line](https://github.com/hoodiehq/hoodie.js/blob/c8ab93dbacab9706a3b43dc37edc12dbc0f202db/src/hoodie/store/localstore.js#L112) passes multiple arguments to resolve. AFAIK this isn't part of any promise spec, and different implementations might have different behaviour. Would be better to pass/return either a 2-item array or a result object like `{ object: object, isNew: isNew }`

janv:
Another pass at that method that would behave correctly:

``` js
defer = getDefer();

try {
  object = helpers.cache(state, object.type, object.id, object, options);
  event = isNew ? 'add' : 'update';
  if (!options.silent) {
    helpers.triggerEvents(state, event, object, options);
  }
  defer.resolve(object, isNew);
} catch (_error) {
  error = _error;
  defer.reject(error.toString());
}

return defer.promise;

```

Just what I said before: move the resolution _after_ the events

grncdr:
hm, so of course right after submitting I realise that maybe you don't _want_ to reject the promise if `triggerEvents` throws. E.g. a bad event handler should not break other things. In that case, `triggerEvents` itself should be wrapping event handlers with `try`/`catch` and expose some mechanism for reporting those errors (e.g. report to `console.error`, emit an `'error'` event, re-throwing on next tick).

I'm not very familiar with hoodie's internal architecture, but to me the right thing to do would be triggering an error event with a default handler that reports stack traces to the console.

gr2m:
@janv 

> Why don't you just trigger the events and resolve the promise afterwards?

The problem I see here is that the promise of something like `hoodie.store.add` would be rejected, although the object was successfully added. I find it very confusing that my external handlers can cause that, it leads to hard to debug issues. That's why I'd prefer to decouple it.

janv:
> I find it very confusing that my external handlers can cause that, it leads to hard to debug issues.

If these external event handlers are out of your control and should not _have_ the ability to alter control flow, their exceptions should not bubble up but be caught inside the event library instead.

If you _want_ to give event handlers the ability to alter control flow, use something like `preventDefault()` which makes that explicit (not sure if that's available in your Event Library).

gr2m:
@grncdr

> One thing I'm trying to figure out is why [this line](https://github.com/hoodiehq/hoodie.js/blob/c8ab93dbacab9706a3b43dc37edc12dbc0f202db/src/hoodie/store/localstore.js#L112) passes multiple arguments to resolve

You are correct, that is a remaining of the time we used jQuery promises. I'll get rid of that, thanks!

gr2m:
Thanks for your inputs! To clarify, when triggerEvents throws, I don't want the promise to be rejected, instead have an alternative way to handle it. The question is, what is the best way to do that. It should be visible to the user out of the box, without the necessity to subscribe to a special `error` event. So I prefer throwing on next tick, or logging to console in some way, either trough `console.error`, or trough an internal `hoodie.log` method, which does not exist yet.

Curious to hear more opinions on it

janv:
With that information I suggest:
Wrap the event triggering code in try/catch and and log exceptions.

grncdr:
> It should be visible to the user out of the box, without the necessity to subscribe to a special error event.

in case it wasn't clear, I was suggesting behavior similar to nodes 'uncaughtException': throw (or log) if nobody handles it, but allow users to explicitly handle it if they want some other behaviour.

We all seem to agree that the try/catch belongs somewhere in or below `triggerEvents` in the call stack though.

gr2m:
I'll merge the workaround with the catch/try as it is better than what we have today, but will make a follow up issue to use `hoodie.log` once it's available, to log uncaught exceptions in event handlers

PR #379, opened 2014-12-21
hoodie.store.updateAll() resolves with array of promises, not the objects

gr2m:
when resolved, enable according tests in https://github.com/hoodiehq/hoodie-integration-test/blob/master/tests/www/api/store.js

see also #380


gr2m:
This is ready for merge. I tested it locally with all CI test

PR #380, opened 2014-12-21
hoodie.store.removeAll() resolves with array of promises, not the objects

gr2m:
when resolved, enable according tests in https://github.com/hoodiehq/hoodie-integration-test/blob/master/tests/www/api/store.js

see also #379


gr2m:
this is ready for review

PR #399, opened 2015-01-02
chore: remove bower & build

### LINKED ISSUE #385: remove /dist folder & build task

lewiscowper:
Think this should fix #385, I think #387 has become way too out of date to merge.

I'm not 100% sure whether the change in index.html is what's needed, but it definitely seems to be.


boennemann:
@lewiscowper imho you can remove index.html (and gh-pages) entirely, as well as cheatsheet.md. All this is covered on the website/docs (where it belongs) :)

lewiscowper:
Including the quickstart?

lewiscowper:
I can't reproduce the failing Travis build here, grunt --verbose test:hoodie passes fine for me. But I've added a couple of JSHint fixes that let the build actually run on Travis.

gr2m:
You can also remove `hoodie.js-assets` from the package.json (https://github.com/hoodiehq/hoodie.js/blob/master/package.json#L44), it was only needed for the index.html file

gabrielmancini:
bower is love... pls fix the bower for all to love u hoodies ;o)

gr2m:
I'll look into this now, a few people run into trouble because the `/dist` files and bower is still floating around.

@gabrielmancini we'll find a good way to make bower users happy, I promise. But for now, it's just causing harm, so I'll get rid of it

svnlto:
@gabrielmancini If you have a running instance of hoodie you can use the following bower.json file

``` js
{
  "name": "yourProject",
  "version": "0.1.0",
  "main": "./app/index.html",
  "dependencies": {
    "hoodie": "http://localhost:[PORT]/_api/_files/hoodie.js"
  }
}
```

whereby [PORT] is the port of your hoodie instance.
This will also make sure you get your plugins bundled up etc.

gr2m:
merged via f76c6c4. Thanks @lewiscowper

PR #407, opened 2015-02-07
WIP: jquery removal

garbas:
**DON'T MERGE (YET)** - _this is work in progress_
- [x] making current tests pass to get the baseline
- [x] pull in pouchdb and setup browserify to use `request-browser` instead of `request` when compiling for browser
- [ ] use `pouchdb/lib/deps/ajax.js` instead of `jQuery.ajax` and use it in `hoodie.request` 
- [ ] check which params are passed to `hoodie.request` across whole hoodie and that there are tests for this cases
- [ ] rebase/cleanup history with feedback from comments on this PR
  since i'm new to hoodie help me and point me to cases which i missed, tnx

[Review on Reviewable](https://reviewable.io/reviews/hoodiehq/hoodie.js/407)


package.json
```diff
@@ -51,5 +54,8 @@
   },
   "scripts": {
     "test": "grunt ci --verbose"
+  },
+  "browser": {
+    "pouchdb/lib/deps/request": "pouchdb/lib/deps/request-browser"
```
svnlto, line 59: :+1:

gr2m, line 59: I wonder, could this eventually break hoodie.js one day, when pouchDB releases a patch version, that moved `pouchdb/lib/deps/request-browser` into an own module? Anything we can do about it?

janl, line 59: hope that SemVer kicks in and they bump their major version, when they do so.

gr2m, line 59: But isn't that internal implementation? From my understanding, it wouldn't require any bumping, it's not something that they document anywhere, it's like using undocumented APIs on other platforms.

But maybe all this is a non-issue :) I just do it in other projects as well, and wondered if there is some common sense on this

svnlto, line 59: if one of your peerDeps changes you make a new release. given the upstream uses semver.

janl, line 59: that’s why I said “hope” :) — we could open an issue with them to consider this as a public API.

janl, line 59: or at least hear what they think about this

nolanlawson, line 59: Agree, that seems pretty fragile. Shouldn't browserify automatically handle it for you, since you've got `pouchdb` in the `dependencies` and we've got [this line](https://github.com/pouchdb/pouchdb/blob/4d135c63efb3128a5e262eca345b451bf717a23e/package.json#L106)?

svnlto, line 59: > Shouldn't browserify automatically handle it for you, since you've got pouchdb in the dependencies

I'm pretty sure, yes.

garbas:
just to ping here, so you'll know i'm still planning to work on this, just current work required all of me last 10days, but there is a "free" weekend coming :)

package.json
garbas, line 59: i'll write some integrational tests for this, then pouchdb changes will be easier to catch.

gr2m:
@garbas thanks for the update Rok, much appreciated :)

gr2m:
you should join forces with https://github.com/hoodiehq/hoodie.js/pull/419

andru:
@garbas I should be able to put a bit of work into this over the coming days. I'll make a fork to try and merge things together. If you get there before me drop me a note :)

PR #411, opened 2015-02-16
fix(account): fix hoodie.account.resetPassword()

### LINKED ISSUE #410: hoodie.account.resetPassword() fails
When invoking `hoodie.account.resetPassword('<some-email>')` I get the following error in the JavaScript console:

```
TypeError: undefined is not a function
    at 11.exports.awaitPasswordResetResult
```

Digging through the source, I tracked down the error to two different lines in https://github.com/hoodiehq/hoodie.js/blob/master/src/hoodie/account/helpers.js

[Line 297](https://github.com/hoodiehq/hoodie.js/blob/master/src/hoodie/account/helpers.js#L297):

``` js
state.events.once('passwordreset', defer.resolve );
```

[Line 302](https://github.com/hoodiehq/hoodie.js/blob/master/src/hoodie/account/helpers.js#L302)

``` js
defer.always( function() {
...
```

It looks like `state.events.once()` should be replaced with `state.events.one()` and `defer.always()` should be replaced with `defer.promise.always()`.

Will send PR shortly.

lupomontero:
In `./src/hoodie/account/helpers.js` replace `state.events.once()` with `state.events.one()` and `defer.always()` replaced with `defer.promise.always()`.

Closes #410

### DIFF
--- a/src/hoodie/account/helpers.js
+++ b/src/hoodie/account/helpers.js
@@ -294,12 +294,12 @@ exports.handlePasswordResetStatusRequestError = function(state, username) {
 exports.awaitPasswordResetResult = function(state) {
   var defer = getDefer();
 
-  state.events.once('passwordreset', defer.resolve );
+  state.events.one('passwordreset', defer.resolve );
   state.events.on('error:passwordreset', exports.removePasswordResetObject.bind(null, state));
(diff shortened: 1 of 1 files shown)


gr2m:
Thanks Lupo!

PR #419, opened 2015-03-30
Remove jQuery as a global dependency.

andru:
This implements (most of) #9.  I've tested it in an application context and as far as I've been able to see it's working as before, only without jQuery!  I haven't been able to do anything on the tests (see below) but I'm submitting this at the suggestion of @janl 

Notes:
- pouchdb-ajax doesn't expose the xhr instance, so some of the stuff that was being done in `request.handleRequestError` is a bit trickier than it was. This pull request introduces one regression: this code no longer has a means to detect if a request errored out because it was aborted. `pouchdb-ajax` might expose this information somehow, but if it does I'm not seeing it.
- I've included lodash, but when bundling with webpack/browserify it will only pull in the two modules I'm actually using - `isEmpty` and `isArray`, which replaces functionality jQuery was providing - so it shouldn't any more weight than an internal dependency for the same functionality would

I'll add some more notes inline.

With regard to tests: I can't get the test runner working, so I've not comitted the little work I've done on them. I've tried running both the full integration tests and the `karma:dev` runner, but neither are getting me anywhere. I get close with `karma:dev` but first up it moans about an invalid `instanceof` in the `function.bind` polyfill (I've googled and I'm not alone. _sidenote: i've got phantomjs 2.0.0 installed but when running `karma:dev` it reports 1.9.8 ... no. idea. how._) and when I find a polyfil it's happy with it has some issues with the code (not the parts I've been working on) but since it won't give me line numbers, and it's already been browserified into a megafile by that point, I'm lost for a path forwards. I'd love some advice here if anyone has got some!

As for implementing the tests, we don't have a global object to stub out anymore, as was being done with `global.jQuery.ajax`, so unless someone has a trick I'm not thinking of we'll need to use something like https://github.com/thlorenz/proxyquireify to stub out pouchdb-ajax.


src/hoodie/request.js
```diff
@@ -64,46 +66,48 @@ exports.request = function(hoodie, type, url, options) {
...
-  return pipedPromise;
+  var ajaxRequest = ajax(extend(defaults, options), function(err, response, data){
+    if(err){
+      return requestDefer.reject(err);
+    }
+    requestDefer.resolve(response, data);
+  });
+  var promise = requestPromise.catch(exports.handleRequestError.bind(null, hoodie));
+  
+  promise.abort = ajaxRequest.abort;
```
andru, line 85: It seems to me like an anti-pattern to attach `abort` to the promise, but I didn't want to take on too much in this first contribution, so I left it be. 
The only place I see it being used is in `account/helpers.js:withPreviousRequestsAborted` - maybe instead of returning the abort method it should be added to a global queue, or maybe this method should return an object:

```
{
promise: promise,
abort: ajaxRequest.abort
}
```

Thoughts?

src/hoodie/request.js
```diff
@@ -64,46 +66,48 @@ exports.request = function(hoodie, type, url, options) {
...
+  //directly, and afaik it doesn't pass along any
+  //means to direct an abort
+  /*if (xhr.statusText === 'abort') {
 
     return rejectWith({
       name: 'HoodieConnectionAbortError',
       message: 'Request has been aborted',
     });
-  }
+  }*/
```
andru, line 107: I can't see any way to get the actual xhr instance from pouchdb-ajax, so I'm not sure how to replicate this functionality

src/hoodie/request.js
```diff
@@ -145,6 +152,14 @@ exports.HTTP_STATUS_ERROR_MAP = {
...
+
+/**
+ * pouch-ajax already parses the error and doesn't pass
+ * along the xhr instance for us to be able to parse the
+ * error here ourselves
+ * so I'm not quite sure how to port this functionality over
+ * without losing some of it
+ */
 exports.parseErrorFromResponse = function(xhr) {
   var error = JSON.parse(xhr.responseText);
```
andru, line 164: As above. `xhr` isn't exposed, and `pouchdb-ajax` already formats the error, so this function is redundant right now, I've just left it in for discussion.

src/hoodie/request.js
```diff
@@ -43,14 +42,17 @@ var exports = module.exports = function(hoodie) {
...
 var API_PATH = '/_api';
 
-exports.request = function(hoodie, type, url, options) {
+exports.request = function(hoodie, method, url, options) {
+  //passing in an XHR instance just lets us test it
   var defaults = {
-    type: type,
-    dataType: 'json'
+    method: method,
+    json: true
```
gr2m, line 49: What happens, if I want to send a request that is not content-type: application/json? e.g. the `POST /_session` request from `account.signIn(username, password)` is such a request. Does it still work for you? It should send `Content-Type:application/x-www-form-urlencoded; charset=UTF-8`, could you please double check?

andru, line 49: Sign in works fine for me. I'll investigate how! :)

src/hoodie/request.js
gr2m, line 85: I'd like to have `.abort` as a standard Promise decoration in Hoodie. I'd also like to add it to other things, like task promises from `hoodie.task.start()` etc. PouchDB has a similar concept with `.cancel` on changes feeds.

So I'd just leave it as is for now

src/hoodie/request.js
gr2m, line 107: Would still be good to differentiate errors coming from manually aborted requests. Any idea? Maybe we can reach out to the PouchDB folks?

andru, line 107: See https://github.com/pouchdb/ajax/issues/1 

Since it's not currently being used by PouchDB proper, I'm not sure how battle-ready the `pouchdb-ajax` library is.  Might be a good idea to look into an alternative.  thoughts?

gr2m:
we should join forces with #407  :)

simoneduca:
@boennemann Why has this been closed? Has jQuery been removed?

gr2m:
Yes, in the upcoming Hoodie release

simoneduca:
👍

simoneduca:
@gr2m when is that due? Link? thanks

gr2m:
no due date. But here are our milestones, we are nearly done with the code part:
http://gr2m.github.io/milestones/

We have a new example app that you can use as a starting point to play with it:
https://github.com/hoodiehq/hoodie-app-something-tracker

PR #424, opened 2015-06-22
forcing GZIP for bootstrapping #423

### LINKED ISSUE #423: Force Gzip
This is the frontend counterpart to https://github.com/hoodiehq/hoodie-server/pull/349.

Before I can prepare a PR,  I need some guidance with how this fits into hoodie.js best.

First off, I think this should _not_ be a plugin, as this is useful for all apps and should be built-in.

I also acknowledge that this code does DOM manipulation, which hoodie.js (afaik) does’t do yet, so we need to make sure this is absolutely fail-safe.

If we can’t make it fail-safe, we need to wrap this in a `hoodie.enableForcedGzip()` method that developers have to call themselves (or with `new Hoodie({force_gzip: true})` or something, it doesn’t matter at this point :).

This is the code that we need to be running:

``` javascript
      // Check for a cookie.
      // If absent, set a session cookie saying “compression NOT ok”.
      // Write out an iframe element to the page.
      if(document.cookie.indexOf('hoodie_force_gzip') === -1) {
        document.cookie = 'hoodie_force_gzip=false';
        var bodyTag = document.getElementsByTagName('body')[0];
        var insertingIframe = document.createElement('iframe');
        insertingIframe.src = '/_api/_files/force-gzip.html?gzip=force';
        bodyTag.appendChild(insertingIframe);
      }
```

More background here: http://www.stevesouders.com/blog/2010/07/12/velocity-forcing-gzip-compression/

cc @gr2m @zoepage

gr2m:
closes #423

### DIFF
--- a/src/lib/store/remote/api.js
+++ b/src/lib/store/remote/api.js
@@ -140,10 +140,23 @@ exports.getSinceNr = function(state) {
 // changes since the beginning, but this behavior might be adjusted,
 // e.g for a filtered bootstrap.
 //
+// As several firewalls and anti virus software strip out Accept-Encoding
+// request headers, ~15% of users end up with much bigger download sizes
+// than needed. So we created a workaround to force gzip via url param,
+// and fallback to normal request if it fails:
+// https://github.com/hoodiehq/hoodie.js/issues/423
+//
+// More background:
+// http://www.stevesouders.com/blog/2010/07/12/velocity-forcing-gzip-compression/
+//
 exports.bootstrap = function(state) {
   state.isBootstrapping = true;
   state.remote.trigger('bootstrap:start');
-  return state.remote.pull()
+  return state.remote.pull({forceGzip: true})
+    .catch(function () {
+      // TODO: make this smarter: restart only if parsing response fails
+      return state.remote.pull();
+    })
     .done(helpers.handleBootstrapSuccess.bind(null, state))
     .fail(helpers.handleBootstrapError.bind(null, state));
 };
@@ -156,8 +169,8 @@ exports.bootstrap = function(state) {
 // We currently make long poll requests, that we manually abort
 // and restart each 25 seconds.
 //
-exports.pull = function(state) {
-  state.pullRequest = state.remote.request('GET', helpers.pullUrl(state));
+exports.pull = function(state, options) {
+  state.pullRequest = state.remote.request('GET', helpers.pullUrl(state, options));
 
   if (state.remote.isConnected()) {
     global.clearTimeout(state.pullRequestTimeout);
--- a/src/lib/store/remote/helpers.js
+++ b/src/lib/store/remote/helpers.js
@@ -174,12 +174,13 @@ exports.mapDocsFromFindAll = function(state, response) {
 // not return a longpoll URL, as we want it to finish right away, even if there
 // are no changes on remote.
 //
-exports.pullUrl = function(state) {
+exports.pullUrl = function(state, options) {
   var since = state.remote.getSinceNr();
+  var forceGzipOption = options && options.forceGzip && '&force_gzip=true' || '';
   if (state.remote.isConnected() && !state.isBootstrapping) {
     return '/_changes?include_docs=true&since=' + since + '&heartbeat=10000&feed=longpoll';
   } else {
-    return '/_changes?include_docs=true&since=' + since;
+    return '/_changes?include_docs=true&since=' + since + forceGzipOption;
   }
 };
 


src/lib/store/remote/api.js
```diff
@@ -140,10 +140,20 @@ exports.getSinceNr = function(state) {
 // changes since the beginning, but this behavior might be adjusted,
 // e.g for a filtered bootstrap.
 //
+// As several firewalls and anti virus software strip out Accept-Encoding
+// request headers, ~15% of users end up with much bigger download sizes
+// than needed. So we created a workaround to force gzip via url param,
+// and fallback to normal request if it fails:
+// https://github.com/hoodiehq/hoodie.js/issues/423
```
janl, line 147: maybe link to the stevesouders post as well

src/lib/store/remote/api.js
```diff
@@ -140,10 +140,20 @@ exports.getSinceNr = function(state) {
...
+//
 exports.bootstrap = function(state) {
   state.isBootstrapping = true;
   state.remote.trigger('bootstrap:start');
-  return state.remote.pull()
+  return state.remote.pull({forceGzip: true})
```
janl, line 155: The option is `{gzip:force}`.

src/lib/store/remote/api.js
```diff
@@ -140,10 +140,20 @@ exports.getSinceNr = function(state) {
...
+    .catch(function () {
+      // make this smarter: restart only if parsing response fails
+      return state.remote.pull()
```
janl: What happens in the scenario:
1. server sends broken JSON, gzipped.
2. parsing fails.
3. server sends broken JSON, ungzipped.

It should be the same as before this patch, right?

src/lib/store/remote/helpers.js
```diff
@@ -174,12 +174,13 @@ exports.mapDocsFromFindAll = function(state, response) {
 // not return a longpoll URL, as we want it to finish right away, even if there
 // are no changes on remote.
 //
-exports.pullUrl = function(state) {
+exports.pullUrl = function(state, options) {
   var since = state.remote.getSinceNr();
+  var forceGzip = options && !!options.forceGzip
```
janl: Again `gzip=force`. My thinking was so that we can do `gzip=off` (force no gzip regardless of headers) and `gzip=headers` (default) later without having to introduce more parameters.

src/lib/store/remote/helpers.js
```diff
@@ -174,12 +174,13 @@ exports.mapDocsFromFindAll = function(state, response) {
...
   if (state.remote.isConnected() && !state.isBootstrapping) {
     return '/_changes?include_docs=true&since=' + since + '&heartbeat=10000&feed=longpoll';
   } else {
-    return '/_changes?include_docs=true&since=' + since;
+    return '/_changes?include_docs=true&since=' + since + '&forceGzip=' + forceGzip;
```
janl: one more `gzip=force`.

src/lib/store/remote/api.js
gr2m, line 155: I think `forceGzip: true` is more clear in terms what it does, don't you think? And I don't see a scenario where we would need something like `gzip: false`

janl, line 155: > And I don't see a scenario where we would need something like gzip: false

See my note below about `gzip: off` and `gzip: headers`.

Debugging comes mind, too.

If this isn’t convincing: I don’t like camel case in url params. Can we do`force_gzip:true` or something similar?

gr2m, line 155: yes, let's do `force_gzip:true` to stay aligned with CouchDB query param name conventions

src/lib/store/remote/api.js
gr2m: yes, result will be the same, only difference being that it now sends two requests instead of one

janl: thanks!

src/lib/store/remote/api.js
janl, line 155: aye, I’ll update hoodie-server

janl, line 155: PR here https://github.com/hoodiehq/hoodie-server/pull/352

gr2m:
fyi I'm debugging this right now, do not merge yet

gr2m:
Okay I've tested it now and it works, together with hoodie-server@4.0.0.

src/lib/store/remote/api.js
gr2m, line 147: done, see below

Issue #458, opened 2016-02-26
simplify Internal Module Architecture


gr2m:
Progress overview, find background info below
- [x] https://github.com/hoodiehq/hoodie-log/issues/31
- [x] https://github.com/hoodiehq/hoodie-connection-status/issues/46
- [x] https://github.com/hoodiehq/hoodie-store/issues/1
- [x] https://github.com/hoodiehq/hoodie-store-server/issues/32
- [x] https://github.com/hoodiehq/hoodie-store-client/issues/90
- [x] create `hoodie-account` repository / module as in https://github.com/hoodiehq/hoodie-store/issues/1
- [x] migrate hoodie-server-account / hoodie-client-account as in https://github.com/hoodiehq/hoodie-store-server/issues/32 and https://github.com/hoodiehq/hoodie-store-client/issues/90

Once all released
- [x] update [hoodie-client package.json](https://github.com/hoodiehq/hoodie-client/blob/65ccb7c346979f76831e7696336179bd67d326c1/package.json#L9-L12)
- [x] update [hoodie-client README.md](https://github.com/hoodiehq/hoodie-client/blob/65ccb7c346979f76831e7696336179bd67d326c1/README.md) (links to repos)
- [x] update [hoodie-client/lib/get-api.js](https://github.com/hoodiehq/hoodie-client/blob/master/lib/get-api.js#L5-L8)
- [x] update `hoodie-server/lib/hapi.js` [here](https://github.com/hoodiehq/hoodie-server/blob/c7e7528415834685356c07573074954ef5b5d8dc/lib/hapi.js#L40) and [here](https://github.com/hoodiehq/hoodie-server/blob/c7e7528415834685356c07573074954ef5b5d8dc/lib/hapi.js#L28)
- [x] update hoodie’s [README.md](https://github.com/hoodiehq/hoodie/blob/tent/README.md#why-is-there-no-code-in-this-repository) with new architecture

---

Today we have this for new Hoodie

```
hoodie/
├── hoodie-client
│   ├── hoodie-client-account
│   ├── hoodie-client-connection-status
│   ├── hoodie-client-log
│   └── hoodie-client-store
└── hoodie-server
    ├── hoodie-server-account
    └── hoodie-server-store
```

Additionally there are these two repositories:
- https://github.com/hoodiehq/hoodie-standalone-account
- https://github.com/hoodiehq/hoodie-standalone-store

We use these to test client & server together and have a simple generic web app UIs.

The problems with this architecture are
[...]

PR #460, opened 2016-04-04
Fix links that were changed in new architecture

finnp:
As part of: https://github.com/hoodiehq/hoodie/issues/458#issuecomment-204634204


gr2m:
LGTM, don’t mind the travis error, it cannot be affected to this as this PR only updates the README

PR #474, opened 2016-05-03
Camp refactoring

gr2m:
Something I worked on over the weekend. This is the last big refactoring for Hoodie’s camp release. Sorry I know it’s a lot of changes, but old integration tests still pass, and it also works with the our end-to-end tests in `app-hoodie-tracker`.

Reviewing might be hard, maybe look through the final code only and see if it makes sense. I’m happy with the result myself


gr2m:
lol no idea why this was closed ?! Can’t even reopen it. Follow up: https://github.com/hoodiehq/hoodie/pull/475

PR #478, opened 2016-05-07
Upgrade to good@7

gr2m:
The upgrade to good 7 needs changes to the way we use `good`, see changelogs: https://github.com/hapijs/good/issues/467

I looked into it but couldn’t figure it out myself, would be nice if someone with more experience good have a glance at it. We only use it in one file: [plugins/logger.js](https://github.com/hoodiehq/hoodie/blob/master/server/plugins/logger.js)


AdriVanHoudt:
for reference: https://github.com/hapijs/good/issues/467

gr2m:
@christophwitzko can you maybe have a look at this? I tried to migrate it but too much stuff changed and we seemed to have used some internal APIs and I have no idea how to migrate it :/

gr2m:
see https://github.com/hapijs/good/issues/500#issuecomment-226604443

gr2m:
closing in favor of https://github.com/hoodiehq/camp/issues/35

PR #482, opened 2016-05-14
Loading of plugin server components

jameswestnz:
@gr2m:

Here's an initial concept for https://github.com/gr2m/milestones/issues/89

A few notes:
- I couldn't see where to get all passed `options` from within `plugins/index.js`, so needed to extend `options` within `config/parse-options.js`. Please let me know if theres a more preferred way.
- I wasn't too keen to use the `package.json` as a means of loading the plugins... thought it was best we use the config in the rc file, as it may be less confusing for users? Open to thoughts
- I haven't created any form of tests yet, so only worth merging in if we're happy it doesn't break general operation of hoodie

Based on this, it's likely not ready for a PR, but is more a means of grouping the changes and discussing.


jameswestnz:
Change coming to this soon - I see why using the RC file may be less ideal... It's more environment specific than app specific, so we need in 'package.json'.

gr2m:
> Change coming to this soon - I see why using the RC file may be less ideal... It's more environment specific than app specific, so we need in 'package.json'.

yes exactly. `package.json` could be app-specific defaults, as it gets checked in to source code. All the `rc` configuration like the `.hoodierc` file is environment specific, so you can run hoodie with a PouchDB backend locally and use a CouchDB in production, for example

jameswestnz:
@gr2m: Got distracted yesterday, but keen to get this closer today!

In regards to package.json structure - old hoodies structure looks like this:
`
  "hoodie": {
    "plugins": [
      "hoodie-plugin-appconfig",
      "hoodie-plugin-email",
      "hoodie-plugin-users",
      "hoodie-plugin-agtract"
    ]
  },
`

I'm proposing we stick to to the `"hoodie": {"plugins": []}` structure, but discard the "hoodie-plugin-" portion. So something like this:
`
  "hoodie": {
    "plugins": [
      "store",
      "account",
      "agtract": {
        "app": "specific"
      }
    ]
  },
`

There's also nothing stopping us still parsing the "hoodie-plugin-" situation, as we'd just `string.replace` on the name prior to doing anything with it.

Assuming that solves "app"-side config, we'll obviously want to have environment config for each plugin. We could either leave the `config/parse-options.js` as I have it (as it simply merges all content from the RC file with the config), or maybe we should be looking to create a `config/plugin-options.js` that accepts the plugin name, and loads from RC/cli flags.

Hopefully this is feeling a little more on track?

gr2m:
it is possible that not all modules which will be used as hoodie plugins will follow the `hoodie-plugin-<name>` naming convention, but most of them will. I’m good with the `"plugins"` key, make things more clear. I’m not yet sure if an array or an object is better though. Probably an array were we can simply list the plugin names, and pass in an object if we want to set custom properties.

So for example, a Hoodie app’s default configuration could look like this:

``` json
{
    "hoodie": {
        "public": "app",
        "account": {
            "sessionTimeout": 1209600
        },
        "plugins": [
            "foo",
            "bar",
            {
                "name": "baz",
                "package": "hoodie-boozie",
                "options": {
                    "bazOption": "value"
                }
            }
        ]
    }
}
```

The `"bazOption"` can also be set using `rc` as `hoodiePluginBazOption`

Makes sense?

jameswestnz:
Need to assume that `options.plugins` will be either an array or object. See https://github.com/hoodiehq/hoodie/pull/486#issuecomment-219410988

If array, it will be converted to an object before bootstrapping plugins.

gr2m:
> Need to assume that options.plugins will be either an array or object. See #486 (comment)

I’d say that in `package.json`, `"hoodie": { "plugins": [] }` is always an array. The values can either be plugin names (strings) or objects. A string is a shortcut for `{"name": "plugin name here"}`.

We would parse these options internally to an object before we merge the config from `rc` into it. 

If that sounds all to confusing, then maybe we should have `hoodie.plugins.*` be an object everywhere, including in the `package.json`, and do something like 

``` json
{
  "hoodie":  {
    "plugins": {
      "pommes": { "secret": 123 },
      "otherplugin": true
    }
  }
}
```

I think eventually we could even support both formats in `package.json`, where the array could be used as shortcut. Would that work?

server/config/parse-options.js
```diff
@@ -4,11 +4,22 @@ var path = require('path')
 
 var defaultsDeep = require('lodash').defaultsDeep
 var log = require('npmlog')
+var extend = require('extend')
```
gr2m: I think you can use lodash’s [defaultsDeep](https://lodash.com/docs#defaultsDeep) instead of `extend`?

server/config/defaults.js
```diff
@@ -27,6 +27,7 @@ function getDefaults () {
     // core modules
     account: {},
     admin: {},
-    store: {}
+    store: {},
+    plugins: []
```
gr2m: so that will be an object now, right? With they keys being the plugin names, and values being either `true` (default values), `false` (disabled) or an object of options

server/plugins/index.js
```diff
@@ -31,7 +32,48 @@ function registerPlugins (server, config, callback) {
       }
     }
   })
-  var plugins = hapiPlugins.concat(localPlugins, hoodieCorePlugins)
+  if (!('plugins' in config)) config.plugins = []
```
gr2m: config.plugins should always be set per https://github.com/hoodiehq/hoodie/pull/482/files#diff-3cc7e17c48506403b02e5e3ae7089a2cR31

jameswestnz:
@gr2m I think I've boiled this down to something more appropriate. Keen on your input!

Plugins can now **only** be enabled in `package.json`. The formats accepted in this file are as follows:

Standard array:

```
{ 
  hoodie: {
    plugins: ["pluginone", "plugintwo"]
  }
}
```

Array of objects

```
{ 
  hoodie: {
    plugins: [{ name: "pluginone" }, { name: "pluginone" }]
  }
}
```

Objects

```
{ 
  hoodie: {
    plugins: { "pluginone": { name: "pluginone" }, "plugintwo":  {name: "pluginone" } } 
  }
}
```

This allows for simple plugin loading, all the way through to defining plugin options as you can see here: https://github.com/agtract/hoodie/blob/10a4e5e84f254c83e510e1b76979b4617089dff3/server/config/parse-options.js#L36-L41

One thing I've changed is how argv/rc options affect plugin loading. This was the thing complicating merging before, and as you mentioned, an environment shouldn't be changing which plugins are loaded, as this should be up to the app configuration. I would suggest if a user wants to be able to do this, we either create an "enabled" option which can be defined in argv/rc, or the user takes advantage of environment variables inside their plugin, detecting what to do. The latter makes more sense to me at present.

So essentially argv/rc options are only used to define a plugins `options` - which means `--plugins.pluginname.secretsalt 1234` maps to a plugins' options and will look similar to `{ name: "pluginone", options: { secretsalt: 1234 } }`

I'm building a plugin at present, so will mention here once it's a working example.

I feel we still need tests for this PR, however maybe this is something we sort pre-stable?

**Note:** Once plugins are passed in either array or object format, they are then converted to an object (a result of running through `defaultsDeep`, where the default is an object), and then mapped back to an array for easy parsing. It felt somewhat dirty doing it this way, however it seems to be the cleanest as it assumes that `package.json` could be in any format, and plugin loading is always from an array of plugins.

server/config/defaults.js
jameswestnz: Might be worth checking my latest comment on this one - slight change from our previous discussion, but worth a look.

gr2m:
sounds all good to me! I’d say let’s see how it works with your plugin? That will help us to see it in an actual real context.

What do you have in mind for the `routes` option here: https://github.com/agtract/hoodie/blob/10a4e5e84f254c83e510e1b76979b4617089dff3/server/config/parse-options.js#L39?

In my thinking, plugins will be similar like the core modules, they will simply be hapi plugins to extend the server, e.g. like the [hodie-account-server plugin](https://github.com/hoodiehq/hoodie-account-server/blob/master/plugin/index.js). So you get the `server` instance and can do whatever you want with it, including defining routes

gr2m:
can you rebase your PR on latest master?

jameswestnz:
@gr2m client loading, including minifying is work. Could be more efficient? I think we need to detect prod or dev environment as to when we minify, as it does slow startup down

gr2m:
> client loading, including minifying is work. Could be more efficient? I think we need to detect prod or dev environment as to when we minify, as it does slow startup down

Yes I agree, but maybe lets do this in a follow up issue?

server/bundle-client.js
```diff
@@ -4,31 +4,63 @@ var fs = require('fs')
...
-  var hoodieClientPath = path.join(hoodieClientModulePath, 'dist/hoodie.js')
+  var hoodieClientPath = path.join(hoodieClientModulePath, 'index.js')
   var bundleTargetPath = path.join(config.paths.data, 'client.js')
-
-  // https://github.com/hoodiehq/hoodie-client/issues/34
-  // var hoodieMinPath = path.join(hoodieClientModulePath, 'dist/hoodie.min.js')
-
+  var bundleTargetMinPath = path.join(config.paths.data, 'client.min.js')
+  
+  // james
```
gr2m: james? :)

server/config/defaults.js
gr2m: I'm sorry I could figure this out by myself I’m just busy conferencing right now, I'll have more time again on Friday - Monday. 

Can you explain why we set `plugins: []` instead of `plugins: {}`? Isn’t the final internal format an object for `config.plugins`? Then we wouldn’t need this in `parse-options.js` I think:

``` diff
 +  // ensure we have all required options
+  _.defaultsDeep(options, {
+    plugins: {}
+  })
```

And instead of `var plugins = Object.keys(pkg.hoodie.plugins).map(function (key) {` we would set it to an object, with the plugin names being the keys?

Am I missing something?

server/bundle-client.js
jameswestnz: Computer obviously wants my name there ;)

Gone now!

server/config/defaults.js
jameswestnz: Have pushed code updating this, an object definitely makes more sense.

jameswestnz:
@gr2m latest commit allows plugins to get access to hoodie config through `server.settings.app`

jameswestnz:
@gr2m here's a working plugin example: https://github.com/agtract/hoodie-plugin-bundle-client

Uses an optional config option `minify` to test the ability to pass options, and also shows the need for `server.settings.app`

gr2m:
Can you add me to https://github.com/agtract/hoodie/ for simpler collaboration on this PR?

gr2m:
Added a few changes: https://github.com/agtract/hoodie/compare/plugins...hoodiehq:agtract-plugins

gr2m:
Things looking good to me, let’s add a few more tests with options in `package.json`, and overrides with `rc`, then let’s ship it :)

jameswestnz:
@gr2m sorry for the delays here - had a crazy past week!

I've merged in your code, all looks good. I'll run some tests against my setup, as I have a few plugins already :)

I'll see if I can get some basic tests in there today too.

jameswestnz:
As a note for anyone that may stumble across this:

I didn't want to create (yet) another repo/npm module for code that was specific only to my hoodie app. I still wanted this logic in a plugin to take advantage of what we've built above. So I decided to store this plugin logic in "./src/plugins/agtract", then link this directory using npm link. Example of what you'd see in `package.json`:

```
  "scripts": {
    "postinstall": "npm link ./src/plugins/agtract/;",
    "postupdate": "npm link ./src/plugins/agtract/;",
  },
```

Which means `require('hoodie-agtract-plugin')` would still work providing that directory had a `package.json` file.

The other reason for doing this was that I've split my server and client logic into two as well - so I have `agtract-server` and `agtract-client` which are dependencies for `hoodie-plugin-agtract` - all the later does is construct the hoodie plugin, the require the other two. Much like Hoodie's account & store plugins.

gr2m:
Hey @jameswestnz, we just shipped app-specific plugins \o/ #637. Check the README here: https://github.com/hoodiehq/hoodie/tree/v25.1.0#extending-hoodie-for-your-app

The implementation is really neat, and I think we are nearly done with 3rd party plugins now. All we have to discuss / figure out is how to configure the plugins that we want to load from npm packages. 

Loading plugins for the server would be added to https://github.com/hoodiehq/hoodie/blob/master/server/plugins/index.js. Loading client plugins would be added to https://github.com/hoodiehq/hoodie/blob/5ec889b2b8f0ce0c3337eced8d0fc871f84d8eda/server/plugins/client/bundle.js#L31-L33

Also see the discussion at https://github.com/hoodiehq/discussion/issues/107 to move the configuration out of the `package.json` to be more flexible, e.g. to have configuration based on different environments, and to pass functions as values etc. would to hear your thoughts!

But independent of that discussion, I would suggest we simply add a configuration option `"plugins"` which is an array of npm package names that are installed and meant to be loaded as plugins. We can then check for `<npm package name>/hoodie/server`  and `<npm package name>/hoodie/client` on-the-fly as we do for the app-specific plugins implemented in #637.

Oh and if `hoodie/public` exists that we should serve the assets at `<hoodie url>/hoodie/<plugin name>/*`, so that plugins can serve their own UI :) But we can do that in a follow up PR if turns out to be tricky

Getting close :)

janl:
Hi, we’re sorry we didn’t get this over the finish line. This is now being tackled again in https://github.com/hoodiehq/hoodie/pull/711 — closing for now, but thanks all for the contributions here <3

PR #486, opened 2016-05-16
Improved argv/environment options

jameswestnz:
**Note:** PR has evolved. Please read thread for further detail.

@gr2m: Wasn't as easy as removing the mapping unfortunately... The `_.mapKeys` there to pass the argv options to `rc` so `rc` _doesn't_ have to do the work. Removing this function made the argv code somewhat unneeded, and also broke any situations that may have expected hyphenated options (such as log level) to not work, as they're converted to camel case there too.

Instead, I thought it's still be best for us to support enabling or disabling plugins from argv. This PR parses the argv options prior to being processed and merged with `rc` options, convering all dot-string options to objects/arrays. As an example:

`--no-plugins` becomes `plugins: false`;
`--plugins[0]="pluginname"` becomes `plugins: [ 'pluginname' ]`;
`--plugins[0].name="pluginname"` becomes `plugins: [ { name: 'pluginname' } ]`;

**The purpose of this PR is to simply "restore" `rc`'s default ability to parse argv dot options.** There may be a better way, so opening up for discussion.

**Note:** Argv options are not merged, so any plugins enabled in your rc files would be disabled but passing `--plugins` at present.


gr2m:
Thanks @jameswestnz. One thing I’m not sure about is whether `--plugins[0]="pluginname"` is a good idea. I think it’s nice to have `--no-plugins` to temporarily disable all plugins, but to pass configurations we should rather have something like `--plugins.pluginname.foo=bar` or similar. For example [optimist supports a dot notation](https://github.com/substack/node-optimist#dot-notation). 

I know that makes things a bit complicated to implement, as the structure of the configuration in package.json diverges from the cli / env arguments, but that’s what we do here at Hoodie :) We come up with the best possible user experience first, and then make it work backwards. 

What do you think?

jameswestnz:
Forgot to mention that '--plugins.pluginname.foo=bar' does work, however the resulting 'plugin' option is an object rather than an array - so I'll either need to convert, or go with 'plugins' being an object. Either way isn't a major change.

I was considering an alternate argv, but was keen too not change too much if not required. I was looking at moving to minimist as 'rc' uses it... So we've got a few options

gr2m:
> Forgot to mention that '--plugins.pluginname.foo=bar' does work, however the resulting 'plugin' option is an object rather than an array - so I'll either need to convert, or go with 'plugins' being an object. Either way isn't a major change.

Maybe we can convert the array of `options.plugins` from `package.json` into an object, e.g. `['plugin1', {name: 'plugin2', options: {foo: 'bar'}}` into `{plugin1: true, plugin2: {foo: bar}`

I can imagine that most of the times plugin will not require custom options and we would just end up with an array of names in `package.json` which is very concise. But internally we would have an object to easily access its setting, or to overwrite / set plugin options via `rc`

jameswestnz:
@gr2m have developed this a bit further.

The original "want" was to have the ability to provide `--plugins.name.option.secret "secret here"` on `npm start`, allowing the user to define which plugins are loaded, or setting their environment information.

This evolved somewhat, as `nopt` didn't support dot notation, and `rc` did. Seemed silly to not take advantage of this, and figured it was an opportunity to consider the best way to load environment variables. Keep in mind this PR is not trying to solve app-level options, which _should_ be defined in `package.json` - this is specifically environmental options. Think API keys are a good example.

The outcome was to change from `nopt` to `yargs`. Reasons:
- `yargs` support dot notation
- help, version, and requirements are for the most part built in - no more `readme.txt` for `npm start --help`
- `yargs` has a bunch of ways to define defaults, whether from a `package.json`, passed object, or a custom function.
- It still supports our camelCase naming convention in the docs, so no need to change any deeper files

**What does this PR give us?**
- easier help file for cli
- dot notation for cli args
- ability to override deep options from rc file

jameswestnz:
Example: `--plugins.first false --plugins.second.options.secret 123456` disables plugin named "first", and sets an option for plugin "second"

bin/start.js
```diff
@@ -43,50 +41,66 @@ if (semver.lt(process.versions.node, '4.0.0')) {
...
+    default: true
+  },
+  data: {
+    type: 'string',
+    default: path.join(process.cwd(), '.hoodie'),
+    normalize: true
+  },
+  'db-url': {
+    type: 'string',
+    default: null
```
gr2m: is default required? Or can we leave it out? I think if we leave out `defualt:` it gets no listed in `--help`, but we can do `default: undefined`

bin/start.js
```diff
@@ -43,50 +41,66 @@ if (semver.lt(process.versions.node, '4.0.0')) {
...
+    type: 'number',
+    default: 6004
+  },
+  'bind-address': {
+    type: 'string',
+    default: '0.0.0.0'
+  },
+  public: {
+    type: 'string',
+    default: path.join(process.cwd(), 'public'),
```
gr2m: would be nice if we could show the relative path here `./public` instead of sth like `/Users/gregor/Projects/my-app/.hoodie`

jameswestnz: Is this in regard to how it shows in `--help`? Agreed, I'll take a look.

gr2m:
Great work! Besides the minor comments, I’m not sure about `--plugins`. I don’t think you should be able to enable/disable plugins per CLI. It’s not a setting that would differentiate in development and production, it’s app-specific. Whether a plugin shall be loaded or not should be exclusively configured in `package.json` I think.

But what we should be able to do is to override plugin-specif options. So if my plugin "pommes" has an option "secret", I’d like to be able to set it "123" via

```
CLI: --plugins.pommes.secret 123
ENV: hoodie_plugins__pomes__secret=123
.hoodierc: { plugins: { pommes: {secret: 123 } } }
```

does that makes sense?

jameswestnz:
@gr2m I actually agree about disabling plugins - it was an unintended feature really, as passing `--plugins.name false` would mean we no longer have info to parse for that plugin. So I guess it's not something we'd document as it'd be confusing, but trying to develop that "feature" out may not provide any benefit?

gr2m:
yeah I don’t mind to allow for `--plugins.name false` if it’s making the implementation simpler, but I wouldn’t document it :)

jameswestnz:
I can confirm that the following works:

```
CLI: --plugins.pommes.secret 123
ENV: hoodie_plugins__pomes__secret=123
.hoodierc: { plugins: { pommes: {secret: 123 } } }
```

bin/start.js
```diff
@@ -97,11 +110,15 @@ if (argv.version) {
...
-}))
+})
+.alias('v', 'version')
+.env('hoodie')
+.epilogue('Options can also be specified as environment variables (prefixed with "hoodie_") or inside a ".hoodierc" file (json or ini).')
+.wrap(Math.min(150, yargs.terminalWidth()))
+.argv
+
+// merge with rc config
+_.defaultsDeep(options, _.omit(rc('hoodie', {}, null), ['config', 'configs']))
```
gr2m: can you explain the `_.omit` call here? Where would `.config` or `.configs` come from?

jameswestnz: they're generated by `rc` - used as a means of knowing where the compiled options came from, however I think retaining these in the options will become misleading to users.

gr2m: agree. Could you add a comment about them? That they are coming from `rc`?

bin/start.js
```diff
@@ -43,50 +40,66 @@ if (semver.lt(process.versions.node, '4.0.0')) {
...
+    type: 'number',
+    default: 6004
+  },
+  'bind-address': {
+    type: 'string',
+    default: '0.0.0.0'
+  },
+  public: {
+    type: 'string',
+    default: undefined,
```
gr2m: I think it would be good to say that default values for `--public` and `--data` are `./public` and `./.hoodie` respectively. We could also add descriptions to them, as well as `--in-memory` and `--db-url`, see https://github.com/hoodiehq/hoodie#usage.

Sorry that I only bring this up now

gr2m:
okay last comment https://github.com/hoodiehq/hoodie/pull/486#discussion_r63503009 – then :shipit:

bin/start.js
jameswestnz: No worries, I agree! Have modified, so these should be all good now.

PR #487, opened 2016-05-16
Fixies

gr2m:
- adds https://github.com/hoodiehq/hoodie-admin & https://github.com/hoodiehq/hoodie-admin-client to architecture docs
- serves generic UIs from Hoodie Account/Store
- sets admin password to "secret" when using PouchDB (will create follow up issue)
- log correct address to terminal when started

### DIFF
--- a/server/config/db/pouchdb.js
+++ b/server/config/db/pouchdb.js
@@ -3,6 +3,10 @@ module.exports = pouchDbConfig
 var existsSync = require('fs').existsSync
 var join = require('path').join
 
+// pouchdb-admins is made to be a PouchDB plugin but does not require a db,
+// so can be used standalone
+var Admins = require('pouchdb-admins').admins
+var get = require('lodash/get')
 var jsonfile = require('jsonfile')
 var randomstring = require('randomstring')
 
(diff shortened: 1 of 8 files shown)


gr2m:
ping @hoodiehq/maintainers. This PR includes several fixes in preparation of the [hackathon at OSCON](https://ti.to/hoodie/oscon-meet-the-hoodies). I’ll need to get it merged beforehand, would be great if someone could have a glance at it. So sorry for the rush <3

server/config/db/pouchdb.js
```diff
@@ -14,19 +18,52 @@ function pouchDbConfig (state, callback) {
...
-  callback(null, state.config)
+  if (adminPassword) {
+    return callback(null, state.config)
+  }
+
+  var admins = new Admins({
+    secret: secret
+  })
+
+  admins.set('admin', 'secret')
```
gr2m, line 47: here is where we currently set the admin password to "secret", which of course is only a short temporary solution (without it admins cannot sign in at all). I’ll create a follow up issue to set the password per CLI

PR #510, opened 2016-06-16
don’t write files when run in memory

### LINKED ISSUE #6: don’t write files when run in memory

JavaScript, Node.js.
### ❓ The Motivation

Hoodie can run in an "in memory" mode, meaning that it doesn’t persist any data to the hard drive. All data vanishes when the server stops. This is great for testing, and for quick deployments to read-only systems like [Heroku](https://www.heroku.com/) or [now](https://github.com/hoodiehq/hoodie-app-tracker/pull/58).

But even when run in memory, the Hoodie Server still writes `.hoodie/config.json` and `.hoodie/client.js`, which is causing issues in read-only systems and is unexpected behavior.
### :dart: The goal

When started with `npm start -- --inMemory` no files should be written at all. The Hoodie client can be kept in memory after bundling and not written to the hard drive. And the `.hoodie/config.json` file should not be written at all. But if it exists, Add a `log.info("Configuration read from .hoodie/config.json")` (with the right dynamic path.

Currently the `.hoodie/config.json` is written when the server secret is generated, see [ hoodie/server/config/db/pouchdb.js ](https://github.com/hoodiehq/hoodie/blob/c0d8fd7/server/config/db/pouchdb.js#L23-L33). But when run in memory, that doesn’t make any sense as there is no initial data like user accounts. The server secret is used to calculate session IDs, but all existing sessions get invalidated on restart as all accounts vanish. 
### :clipboard: Step by Step
- [x] claim this issue @gr2m
- [x] Implement the feature :) 
- [x] Commit with `feat: don’t write files when run in memory`  
  Don’t worry about wording or amount of commits. We will clean up everything before we merge :) Also please commit work in progress things if you get stuck, we can work on it together in the PR.
- [x] Start a Pull Request: https://github.com/hoodiehq/hoodie/pull/510

gr2m:
I changed the following:
1. instead of bundling the hoodie client on startup, it now gets bundled on the first `GET /hoodie/client.js`
2. Unless run with `inMemory: true`, the bundled client is written to `<data-path>/client.js`. The `mtimes` of `hoodie-client/dist/client.js` and the bundle are compared, and only re-bundled if there is an update
3. We don’t create the `<data-path>` folder if run with `inMemory: true`
4. We don’t wirte `<data-path>/config.json` if run with `inMemory: true`

closes hoodiehq/camp#6


gr2m:
ping @hoodiehq/maintainers 👀 

Once we have the new version release, deploying to read-only targets like Heroku, now or HyperDev for showcases or quick testing will be a breeze

server/plugins/client/bundle.js
```diff
@@ -0,0 +1,66 @@
...
+      callback(null, buffer, hasUpdate)
+    })
+  })
+}
+
+function getModifiedTime (path, callback) {
+  fs.stat(path, function (error, stats) {
+    // we can ignore error, it means the file does not exist which is fine
+    if (error) {
+      return callback(null, -1)
```
HipsterBrown, line 47: Does `-1` represent `false` for the expected callback for this function?

gr2m, line 47: as we want to compare "modified times" of files, I thought `-1` would be the clearest to return, as it will always be smaller than any date. But maybe we can make this more clear?

HipsterBrown, line 47: That makes sense. Maybe we can add a comment above the function that describes what `callback` is expecting for the return arguments, i.e. "callback is expecting an error or null for the first argument and a number to compare the initial time the file was created/updated. -1 declares that the file was never created."

HipsterBrown:
I saw some namespacing changes from `hoodie` to `hoodie-local`. What's the context behind that?

gr2m:
> I saw some namespacing changes from hoodie to hoodie-local. What's the context behind that?

These were changes of the names of the internal hapi plugins. All hapi plugins must have unique names, otherwise an error will be thrown. The internal plugins of `hoodie` did not follow a naming convention or were simply prefixed with `hoodie-` which could cause conflicts in future when people create plugins, so I thought the best idea would be to use the `hoodie-local-` prefix for these

HipsterBrown:
Only had two comments for understanding the context behind some of the changes. Overall, LGTM and I'm very excited to see how this enables folks to test out Hoodie from the browsers! :clap: :rocket:

jameswestnz:
What happens in these memory-only environments if you attempt to write a file? Seems like all cases are covered here, but am wondering if we should be testing (somehow..) that no write attempts are made out of memory in case an error is caused. This is only a testing question, code is great otherwise 👍

gr2m:
> wondering if we should be testing

Yes I had the same thought, but I simply couldn’t come up with a simple way to test it :/

If you are okay with merging, you need to have `LGTM` in your comment, so that the https://lgtm.co/ gives its green light :) Okay for this time because now my own LGTM counts :D just wanted to let you know for the future

jameswestnz:
Sweet as, thought that would be the case. I couldn't come up with an efficient solution either!

I sure am, so LGTM!

gr2m:
it’s out \o/ [v22.5.0](https://github.com/hoodiehq/hoodie/releases/tag/v22.5.0)

PR #520, opened 2016-07-03
Generate client bundle using url from hoodie host and port config

### LINKED ISSUE #519: Wrong hoodie url when using from another host
The client.js bundle initialize hoodie without url options, like this:

``` javascript
hoodie = new Hoodie();
```

This causes hoodie to use the url from `location.origin`, the problem is that in my use case, I need to use hoodie as an external resource, like this:

``` html
<script type="text/javascript" src="http://localhost:4000/hoodie/client.js"></script>
```

But my application is running on another port (e.g. localhost:8000), so hoodie does not work. And if I instantiate it like this on my app side:

``` javascript
hoodie = new Hoodie({url: 'http://localhost:4000'});
```

It's too late because `client.js` already initialized hoodie on the wrong url, so some automatic calls, like `api/user%2Fmyuser/`, keep getting called on my own app url (localhost:8000/hoodie instead of localhost:4000/hoodie like it should be), throwing 404s, and poor clueless PouchDB says they are normal

rogeriochaves:
This would fix #519

The only problem is the hardcoded protocol (http), it would be better if Hoodie constructor accepted an protocol-agnostic url like this `new Hoodie('//127.0.0.1:8080')`, but currently it doesn't work, could be tackled on a future PR

Another solution (which I prefer, but would have more impact) would be stop instantiating hoodie on `client.js`, and let users do the `new Hoodie(opts)`

### DIFF
--- a/server/plugins/client/bundle.js
+++ b/server/plugins/client/bundle.js
@@ -3,6 +3,7 @@ module.exports = bundleClient
 var fs = require('fs')
 
 var parallel = require('async').parallel
+var defaultsDeep = require('lodash').defaultsDeep
 
 /**
  * we compare the mtime (modified time) for the Hoodie client and the target
@@ -58,6 +59,13 @@ function buildBundle (config, hoodieClientPath, callback) {
       return callback(error)
     }
 
+    if (config.connection) {
+      defaultsDeep(config, {
+        client: {
+          url: 'http://' + config.connection.host + ':' + config.connection.port
+        }
+      })
+    }
     var options = config.client ? JSON.stringify(config.client) : ''
     var initBuffer = Buffer('\n\nhoodie = new Hoodie(' + options + ')')
 
--- a/test/unit/plugins-client-bundle-test.js
+++ b/test/unit/plugins-client-bundle-test.js
@@ -37,12 +37,17 @@ test('bundle client', function (group) {
       }
     })
 
-    bundleClient('client.js', 'bundle.js', {}, function (error, buffer) {
+    bundleClient('client.js', 'bundle.js', {
+      connection: {
+        host: '127.0.0.1',
+        port: 8080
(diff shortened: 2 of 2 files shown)


gr2m:
You can use the `client.url` config for that, but I agree we could default that value to `'http://' + config.connection.host + ':' + config.connection.port` unless `client.url` is set.

Would that make sense? If yes, could you adapt your PR? I would suggest to add another test to `plugins-client-bundle-test.js`

rogeriochaves:
Yeah but right now I can't set this config also, it isn't picked up even if I pass --config.url when starting hoodie

gr2m:
> Yeah but right now I can't set this config also, it isn't picked up even if I pass --config.url when starting hoodie

What Hoodie version do you have currently installed in your app? `npm ls hoodie` will tell you

gr2m:
I’m sorry, it doesn’t seem to work, just tried it myself. Sorry for the confusion. I thought i had it working before ... I’m looking into it

gr2m:
Okay I fear it never got implemented properly. We currently combine the `rc` options with options from the package.json (on the `"hoodie"` key), but we don’t set `config.client` from the options. I’ll look into this today

Your suggestion to default the client URL to `config.connection` will work on your machine, but once you deploy it the internal IP address will not work, the URL will rather be something like `https://myapp.com` than `0.0.0.0:8080`. So this could also break applications, because if no URL is passed, it will connect to the same host which will always work if you load assets from the the same host, while setting it to `config.connection` will break it.

rogeriochaves:
Yes, you are correct, I haven't thought about this, so instead we should fix reading the client options, I'll close this PR then 👍

gr2m:
this one should fix the problem: https://github.com/hoodiehq/hoodie/issues/522

PR #522, opened 2016-07-03
options.url

### LINKED ISSUE #521: add url option
The `url` option is the public URL until which the Hoodie server can be reached, so by default, when run locally, it is http://127.0.0.1:8080

The `url` option will be passed to the client. This way you can load the Hoodie client from any host and it will always connect to the hoodie server. We will also set `config.url` so that plugins can use the public URL as they need it.

gr2m:
Set `config.url` from `options.url` and pass it to the client bundle. Follow up for https://github.com/hoodiehq/hoodie/pull/520#issuecomment-230170420.

closes #521

### DIFF
--- a/README.md
+++ b/README.md
@@ -78,13 +78,13 @@ App-specific default values can be set in `package.json` in `"hoodie": {}`, e.g.
 
 option        | default       | description
 ------------- | ------------- | -------------
+bindAddress   | `'127.0.0.1'` | Address that Hoodie binds to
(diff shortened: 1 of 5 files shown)


gr2m:
ping @hoodiehq/maintainers

rogeriochaves:
LGTM 🚢 🇮🇹

gr2m:
I’m confident about this one and it’s blocking @rogeriochaves, so going to self-LGTM :) 🚢

PR #529, opened 2016-07-16
hapi plugin

### LINKED ISSUE #528: Turn hoodie into hapi plugin
at the moment, `require('hoodie')` returns a method that creates a Hapi instance and passes it to the callback: https://github.com/hoodiehq/hoodie/blob/master/server/index.js

The way we start the Hoodie server at the moment can be seen in [bin/start.js](https://github.com/hoodiehq/hoodie/blob/master/bin/start.js#L129-L145)

I think it had historical reasons why we ended up with that architecture, and I don’t think it makes much sense any more. Instead, I’d suggest that `require('hoodie')` will return a hapi plugin and that we instantiate a Hapi instance in `bin/start.js` instead. 

That will also make it simple to use Hoodie in existing architectures. And for some people who don’t want to wait for plugins, they can build their own hapi server with custom routes and add hoodie to the mix.

Any thoughts?

gr2m:
# WIP

TODOS
- [x] split logic / options between CLI and hapi plugin
- [x] document usage for both CLI and hapi plugin
- [x] fix: the order of how config is read out right now is not correct. Defaults set for CLI are currently overriding configurations set in other places
- [ ] move hapi plugin logic and usage docs to `@hoodie/server` module, [see my comment below](https://github.com/hoodiehq/hoodie/pull/529#issuecomment-233106277). This is just an note to self, I will create a follow up issue for further discussion, it is out of scope for this PR

close #528


gr2m:
when the hoodie server part becomes a hapi plugin, the separation into `hoodie` and `@hoodie/server` makes sense again. `hoodie` would start the hapi server, register the `@hoodie/server` hapi plugin and add the routes for the bundled hoodie client. That way `@hoodie/server` does not need `@hoodie/client` as dependency, which was the main problem why we combined `hoodie` and `@hoodie/server` a few weeks ago

jameswestnz:
Making a lot more sense now - good work!

gr2m:
this is ready for review @hoodiehq/maintainers

rogeriochaves:
With this change hoodie can be added to [hapi's plugin page](http://hapijs.com/plugins) 👍

README.md
```diff
@@ -71,24 +71,104 @@ You can find a more thorough description in our [Getting Started Guide](http://d
...
+}
+```
+
+The priority of configuration:
+
+1. command line arguments
+2. Environment variables
+3. `.hoodierc` files
+4. Your app’s defaults form `"hoodie"` key in `"package.json"`
+5. Hoodie’s defaults as shown in table above
```
rogeriochaves, line 121: 👍

cli/parse-options.js
```diff
@@ -24,23 +21,16 @@ var getDefaults = require('./defaults')
...
   var config = {
     loglevel: options.loglevel,
     paths: {
       data: options.data,
       public: options.public
     },
-    connection: {
-      host: options.bindAddress,
-      port: options.port
-    }
```
rogeriochaves, line 39: why were those removed?

test/unit/cli/parse-options-test.js
```diff
@@ -5,17 +5,19 @@ var test = require('tap').test
...
+    hoodie: {
+      name: 'foo',
+      paths: {
+        data: 'data path',
+        public: 'public path'
+      },
+      connection: {
+        host: 'host name',
+        port: 'app port'
+      },
```
rogeriochaves: connection key was removed from the source but not from this test

cli/parse-options.js
gr2m, line 39: here we parse the CLI options to config that we pass to Hoodie. `connection.host` and `connection.port` are only relevant for hapi, not for Hoodie

test/unit/cli/parse-options-test.js
gr2m: good catch will fix

gr2m:
> With this change hoodie can be added to hapi's plugin page 👍

That’s a great idea :) I’d maybe move the hapi plugin logic to `@hoodie/server` and list that one

rogeriochaves:
I've downloaded the branch to test locally and I'm using this options:

```
node_modules/hoodie/bin/start.js --admin-password 123 --bindAddress 0.0.0.0 --port 4000 --db-url http://admin:123@127.0.0.1:5984 --url http://localhost:4000
```

And because of the `--url` option I'm getting:

```
[1] "host" must be a valid hostname
```

If I remove the port (`:4000`) it works, but I need to set the port for my case.

gr2m:
@rogeriochaves fixed

christophwitzko:
LGTM 👍

gr2m:
Thanks for your help shipping this 👍 🎉

PR #536, opened 2016-07-21
move PouchDB defaults config from CLI to server logic

### LINKED ISSUE #535: move PouchDB defaults config from CLI to server logic
when I use `hoodie` as a hapi plugin, it creates all databases in the root path. The reason is that `config.db.prefix` does not get set, as the logic currently lives in the CLI logic which makes little sense. We have to move it to the server

gr2m:
fixes  #535
ping ping @hoodiehq/maintainers

### DIFF
--- a/cli/parse-options.js
+++ b/cli/parse-options.js
@@ -1,9 +1,6 @@
 module.exports = parseOptions
 
-var path = require('path')
-
 var log = require('npmlog')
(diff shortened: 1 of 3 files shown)


jameswestnz:
Was just about to submit an issue for this! Thanks for spotting so quick.

LGTM!

gr2m:
🏃💨 LGTM

Issue #532, opened 2016-07-20
/hoodie/client.js browserify on demand


gr2m:
We are using `@hoodie/client`’s dist build and then only append a line like `window.hoodie = new Hoodie(options)`. Which is great because it’s fast and doesn’t need any browserify. But there are two problems with this approach:
1. If there is a new release with any client library like [@hoodie/account-client](https://github.com/hoodiehq/hoodie-account-client) then we need to trigger a release of `@hoodie/client` too, because the `/dist` folder gets updated on release only
2. As we will support 3rd party plugins soon:tm:, we will have to browserify on the fly to benefit from npm’s deduplication of dependencies. 

For an example, see how we browserify the client in the [@hoodie/account dev server](https://github.com/hoodiehq/hoodie-account/blob/master/example/index.js#L11-L24)

PR #538, opened 2016-07-24
feat: use browserify to bundle hoodie client (#532)

sdvg:
This makes the `/hoodie/client.js` request use Browserify to bundle an up to date version of @hoodie/account-client instead of delivering the prebuilt version in `/dist`.

For more details see issue #532.

### DIFF
--- a/package.json
+++ b/package.json
@@ -17,6 +17,7 @@
     "@hoodie/store": "^1.0.1",
     "async": "^2.0.0",
     "boom": "^3.1.2",
+    "browserify": "^13.0.1",
     "good": "^6.6.3",
     "good-squeeze": "^4.0.0",
     "h2o2": "^5.1.0",
--- a/server/plugins/client/bundle.js
+++ b/server/plugins/client/bundle.js
@@ -1,7 +1,7 @@
 module.exports = bundleClient
 
+var EventEmitter = require('events').EventEmitter
 var fs = require('fs')
-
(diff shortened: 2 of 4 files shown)


gr2m:
LGTM, great work 👍. I’ve checked it locally with `hoodie-app-tracker` and its integration tests and they all pass.

Ping @hoodiehq/maintainers for 2nd 👀

HipsterBrown:
Is this a breaking change since it changes the `hoodieClientPath`?

Otherwise, LGTM 👍

varjmes:
Once @HipsterBrown has his question answered, this LGTM. Thanks @sdvg! ✨

rogeriochaves:
LGTM

gr2m:
@HipsterBrown @Charlotteis 

This change is part of the browserify on demand

``` dif
-  var hoodieClientPath = path.join(hoodieClientModulePath, 'dist/hoodie.js')
+  var hoodieClientPath = path.join(hoodieClientModulePath, 'index.js')
```

Instead of using the pre-built file at `dist/hoodie.js`, we now need [the main file of `@hoodie/client`](https://github.com/hoodiehq/hoodie-client/blob/master/index.js) which is `index.js`. So it’s all good, it’s only the input path. The output path remains the same.

Thanks for your reviews everyone, this is great 👏 I’ll go ahead and merge it now

PR #556, opened 2016-08-27
hapi usage without options.path throws

### LINKED ISSUE #554: hapi usage without options.path throws
In my recent changes, I introduced this line:
https://github.com/hoodiehq/hoodie/blob/master/server/index.js#L27

The option defaults are only set later, so this line will throw with `TypeError: Cannot read property 'data' of undefined` unless `options.paths` is an object, see https://github.com/hoodiehq/ember-hoodie/pull/21

I’m looking into it

gr2m:
closes #554 
closes https://github.com/hoodiehq/ember-hoodie/pull/21

### DIFF
--- a/server/index.js
+++ b/server/index.js
@@ -24,6 +24,19 @@ function register (server, options, next) {
       log.info('config', 'Storing all data in memory only')
     } else {
       PouchDB.plugin(require('pouchdb-adapter-leveldb'))
+
+      // this is a temporary workaround until we replace options.db with options.PouchDB:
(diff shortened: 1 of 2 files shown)


varjmes:
LGTM! But looks like coverage decreases a lil?

brunopedroso:
Looks good for me. Would only check this coverage though...

gr2m:
Yeah the coverage is another open issue :) 
https://github.com/hoodiehq/camp/issues/4

PR #559, opened 2016-08-30
fix: don't modify the passed options object

rmehner:
So in this project we're registering multiple servers as to Hapi, kinda like this:

``` js
var options = {
  db: {
    url: process.env.COUCH_URL
  }
}

server.register([
  { register: hoodie, options: options },
  { register: otherEndpoint, options: options }
], function (error) {})
```

the problem is, due to hoodie server's behaviour here, `otherEndpoint` never sees what's in `options.db` and will always have an empty object.

This PR fixes this behaviour by cloning the options object before doing anything on it.

### DIFF
--- a/server/index.js
+++ b/server/index.js
@@ -10,10 +10,12 @@ var corsHeaders = require('hapi-cors-headers')
 var hoodieServer = require('@hoodie/server').register
 var log = require('npmlog')
 var PouchDB = require('pouchdb-core')
+var cloneDeep = require('lodash').cloneDeep
 
 var registerPlugins = require('./plugins')
 
 function register (server, options, next) {
+  options = cloneDeep(options)
   if (!options.db) {
     options.db = {}
   }
--- a/test/integration/server-test.js
+++ b/test/integration/server-test.js
@@ -32,6 +32,19 @@ test('server with options.db.url lacking auth', function (t) {
   })
 })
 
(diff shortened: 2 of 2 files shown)


gr2m:
LGTM, thanks Robin!
ping @hoodiehq/maintainers

server/index.js
```diff
@@ -10,10 +10,12 @@ var corsHeaders = require('hapi-cors-headers')
 var hoodieServer = require('@hoodie/server').register
 var log = require('npmlog')
 var PouchDB = require('pouchdb-core')
+var cloneDeep = require('lodash').cloneDeep
```
janl, line 13: didn’t we decide on full `loadash` on the server instead of single functions?

No blocker tho

janl:
LGTM

server/index.js
rmehner: > didn’t we decide on full loadash on the server instead of single functions?

I've adapted from here: https://github.com/hoodiehq/hoodie/blob/7a989154c766419a1f73c29bc5a2860e94c470aa/cli/options.js#L3

But yeah, found out about the `CODING_STYLE.md` now, will create another PR that fixes this inside of `server/index.js`.

gr2m: this is the full lodash. the alternative would have been be `require('lodash/cloneDeep')`

PR #561, opened 2016-08-30
style(server): require full lodash

rmehner:
This is in accordance to `CODING_STYLE.md`.

See https://github.com/hoodiehq/hoodie/pull/559#discussion_r76828935

### DIFF
--- a/server/index.js
+++ b/server/index.js
@@ -10,12 +10,12 @@ var corsHeaders = require('hapi-cors-headers')
 var hoodieServer = require('@hoodie/server').register
 var log = require('npmlog')
 var PouchDB = require('pouchdb-core')
-var cloneDeep = require('lodash').cloneDeep
+var _ = require('lodash')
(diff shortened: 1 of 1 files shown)


gr2m:
That’s fine I think, we do it all over the server code. The thing we try to do is requiring the entire `lodash` package on server, while requiring the separated packages for the client

rmehner:
> That’s fine I think, we do it all over the server code. The thing we try to do is requiring the entire lodash package on server, while requiring the separated packages for the client

Yep, that's what it says in the CODING_STYLE.md. I mean, `require('lodash').deepClone` requires the full lodash, but I guess that's something beginners might stumble over, so the way in the PR might be clearer and more approachable :)

gr2m:
yeah you are right, I don’t mind merging, LGTM :)

varjmes:
LGTM!

PR #580, opened 2016-09-29
Repl server

gabrielperales:


gabrielperales:
I have to add more tests. Now I'm testing that I can connect to the REPL server, but also I have to test that the admin client is on the REPL scope.

gabrielperales:
I've tried to have those tests in separately, but when I exit from the test I need to execute `process.exit(0)` otherwise the REPL server still being opened  and the test runner keep waiting while the REPL server is being executed, so at the end it finish with a timeout. The problem is that if I have another test I have to open the REPL server on other port, because the other server still listening on that port... So I don't know how to do it better... yet.

gabrielperales:
weird.... why it  is failing on travis?... it is passing on my laptop!

gr2m:
@gabrielperales oh sorry nobody got back to you, but looks like you figured it out? How is it going? Can we help in any way?

gabrielperales:
@gr2m yeah, test were failing in travis because I had configured the connection to the RELP through localhost, and I've seen that I had to remove that... When I fixed that I saw that now the coverage was 100% and I had to do some changes to keep it 100%. Now seems to be everything OK

gr2m:
Can you amend the `README.md` file with info on how to use the console? When we add new features, we usually start with updating the `README.md`, it’s a good way to think about what the ideal end result will look like, and help to stay on track, especially with bigger features like this one :)

gr2m:
I’m not sure how to test it. Is it that I start the app with `npm start -- console` and then I am supposed to see an input at localhost:9000?

Our idea was to not start the server at all, instead show an input right in the terminal, similar to happens when you just type in `node` on your terminal. Does that make sense?

I’m very sorry if there was a misunderstanding?

capellini:
@gr2m @gabrielperales FYI - this PR in it's current form undoes some of the changes made in [this PR](https://github.com/hoodiehq/hoodie/pull/591), where we changed the `.bindAddress` option to `.address`.

gr2m:
@gabrielperales can you please rebase the pull request on latest master? Let me know if you need help with that :)

gabrielperales:
@gr2m yeah, I told you here https://github.com/hoodiehq/camp/issues/7#issuecomment-250314124 that there was two ways to implement that, and go for the server one because there was a hapi plugin managing that thing. 

To connect to the RELP you have to open a telnet session: `telnet localhost 9000`. I know that it could be a awful... but otherwise you will get your logs on the same screen and that is also ugly.

Now I'm a little bit busy, I will update the readme and rebase master ASAP.

gr2m:
Ah gotcha. `telnet localhost 9000` is not a good solution for us, we are looking for something similar to how `rails console` behaves, see https://www.youtube.com/watch?v=CDSRGdH7Lnk&feature=youtu.be&t=1m40s. Does that make sense?

PR #583, opened 2016-10-01
First version of the placeholder UI when no public folder is present.

### LINKED ISSUE #19: What could be a useful UI if an app has no `public/` folder

Hoodie is a generic backend for your application, but can also serve your static html application. The latter is not required, it’s a common use case that people run their Hoodie Server somewhere like https://myhoodieapp.com` and the static app is served directly from GitHub pages or [surge.s](http://surge.sh/) or another hoster for [static apps](https://staticapps.org/). In that case, when opening the URL of the Hoodie Server should show some generic fallback UI.
### :dart: The goal

What could we show in this generic Hoodie UI? Maybe the version? Maybe an explenation of what it is? Maybe link to the other generic UIs at `/hoodie`, `/hoodie/account` and `/hoodie/store`? Maybe we should directly show the admin UI that is usually at `/hoodie/admin`? We are open for ideas

This is not really a bug that needs to be fixed or a feature that needs to be implement, but an invite for a discussion. Please ask all your questions here, we are looking forward for your thoughts! Ping us in the [Hoodie Chat](http://hood.ie/chat/) or on [Twitter](https://twitter.com/hoodiehq/) if you have any questions :)

kouryuu:
closes https://github.com/hoodiehq/camp/issues/19
The interface can be found here http://codepen.io/kouryuu/pen/VKzkqL
I have successfully tested this interface using a demo app and works fine in:
- Chrome 53.0.2785.116
- Safari 10.0
- Firefox 48.02

To test it please create a new hoodie project and run `npm start` without a public folder.


gr2m:
> To test it please create a new hoodie project and run npm start without a public folder.

There is a simpler way to test it: you can run `npm start` right in the `hoodie` repository, then open http://localhost:8080 :)

public/index.html
```diff
@@ -3,24 +3,81 @@
   <head>
     <meta charset="utf-8">
     <title>Hoodie</title>
-  </head>
-  <body>
-    <h1>Hello from Hoodie!</h1>
+    <style>
```
gr2m: Can we move the CSS into an external `style.css` file?

public/index.html
```diff
@@ -3,24 +3,81 @@
...
+    body{
+background-color:#FCFCFC;
+animation-name: graytocolorimg;
+animation-duration: 4s;
```
gr2m: Can you explain your thinking behind the animation? I think I’d prefer to keep it very plain and simple, but maybe I’m missing something?

public/index.html
```diff
@@ -3,24 +3,81 @@
...
+}
+h1{
+  font-family:fantasy,sans-serif;
```
gr2m: I think `fantasy` looks cute, but we don’t really use it anywhere else. I’d suggest to just stick with `Helvetica, Airal, sans-serif`. We can improve this later, once we put more thinking into a coherent Hoodie Design :)

public/index.html
```diff
@@ -3,24 +3,81 @@
   <head>
     <meta charset="utf-8">
     <title>Hoodie</title>
-  </head>
-  <body>
-    <h1>Hello from Hoodie!</h1>
+    <style>
+    body{
+background-color:#FCFCFC;
```
gr2m: Can you keep the indentation coherent? We use 2 spaces for CSS

gr2m (changes requested):
Thanks Rodrigo, this is a lot better than what we have right now :) There are a few things I would like to change, and I have some ideas on how further improve it, but maybe we can do it in a follow up PR, and maybe @distalx has ideas, too?

public/index.html
kouryuu: Sure, I was intending to do it very simple and straight forward so I though an external style sheet was not needed since there are very few styles... but thinking in all the other fallback interfaces we will be doing maybe it makes sense to have it in a separate css.

gr2m: yeah, I’d do the same thing in the other interfaces. 

By the way the UIs for `/hoodie/*` are not fallback interfaces, they are actually features of `hoodie` than cannot be changed :) The fallback interface is only for the `/` root path

public/index.html
kouryuu: I was thinking of a very subtle animation; maybe to catch the user's attention that this is an important message.

public/index.html
kouryuu: I'm on it!

public/index.html
kouryuu: Yeah I was really looking for a different and "fun" typography maybe a styleguide would really be helpful.

gr2m: I’d say let’s focus more on content for now and keep the styling simple until we have a CI?

public/index.html
kouryuu: Yeah but it still makes sense to have them with a uniform style maybe a **global** css where both `/hoodie/*`and the message that shows you there is no `/public`folder have a uniform branding and style.

gr2m: A global CSS is tricky. We tried something like this with https://github.com/hoodiehq/hoodie-css, but it turned out to be a maintainance nightmare.

Let’s think about design guidelines and shared CSS later, I’d like to merge your PR and then iterate

public/index.html
kouryuu: Done.

public/index.html
kouryuu: Done.

public/index.html
kouryuu: I left the animation but made it 2s, to make it even more subtle.

public/index.html
kouryuu: Ok, maybe not totally global for now but at least it would be good to be able to reuse the styling for other UI in hoodie.

kouryuu:
@gr2m does this need anything else to be merged?

public/index.html
```diff
@@ -3,24 +3,28 @@
   <head>
     <meta charset="utf-8">
     <title>Hoodie</title>
+<link rel="stylesheet" type="text/css" href="style.css">
```
varjmes: Can you indent this so it is in line with the rest of the code, i.e the title, please.

public/index.html
```diff
@@ -3,24 +3,28 @@
...
-    <p>This is necessary for your own javascript files to pick up the hoodie methods.</p>
-    <p>Now handle it just like you would handle your frontend project, hoodie won't get in your way. </p>
-    <p>4. Setup your asset folders and create your .css and .js files.</p>
-    <p>5. You’ll need to restart the Hoodie server so it servers assets from your app’s <code>public</code> folder</p>
-
-    <p>If you want to learn more, check the <a href="http://docs.hood.ie/camp/">work-in-progress documentation</a></p>
-    <p>Happy coding :)</p>
-</body>
+      <h2>Get going with Hoodie</h2>
+  <ol>
```
varjmes: looks like this should be in line with the `h2` above and thus needs some indentation.

public/style.css
```diff
@@ -0,0 +1,44 @@
+body {
+  background-color:#FCFCFC;
+  animation-name: graytocolorimg;
+  animation-duration: 2s;
+}
```
varjmes, line 5: may we have a line break between each ruleset please.

i.e there should be lines between each block

``` css
block {
  thing: 'thing';
}

block2 {
  thing: 'thing'
}
```

varjmes (changes requested):

public/index.html
kouryuu: Thanks @Charlotteis I improved the indentation on the index.html on the latest commit

public/style.css
kouryuu: Done. Thanks :)

kouryuu:
@Charlotteis thank you for your feedback, it made the code a whole lot neater.

public/index.html
```diff
@@ -3,24 +3,28 @@
...
+    <link rel="stylesheet" type="text/css" href="style.css">
   </head>
-  <body>
-    <h1>Hello from Hoodie!</h1>
 
-    <p>You see this because your app has no <code>public/</code> folder.</p>
-    <p>But don't worry, we're here to get you started! </p>
+    <body>
+      <h1>Well done, you made it!</h1>
+    <div id="doggy"><img src="low-profile-dog-1.png" /></div>
```
varjmes: indentation is still off here, everything should be indented within the body.

```
body
  h1
  div
  div
```

etc.

varjmes (changes requested):

public/index.html
kouryuu: Ok so if I understand correctly the root element is `<html>`and since `<body>` is its children it should get one identation.

kouryuu: ✅  Done. Also I forgot to close the `<body>` tag my bad.

public/index.html
```diff
@@ -3,24 +3,29 @@
...
-
-    <p>If you want to learn more, check the <a href="http://docs.hood.ie/camp/">work-in-progress documentation</a></p>
-    <p>Happy coding :)</p>
-</body>
+    <h1>Well done, you made it!</h1>
+    <div id="doggy">
+      <img src="low-profile-dog-1.png" />
+    </div>
+    <div class="info">
+      Hoodie is now running a local server.
```
gr2m: can you wrap this line into `<p>...</p>` tags and reduce the indentation of the lines below by two spaces so it’s the same alignment? Or maybe `h2` tags would be better?

gr2m (changes requested):
just a minor from my side, otherwise great

public/index.html
kouryuu: I think `<h2>` makes it way too visible perhaps a `<strong>` tag (although nowadays its being phased out a little). 

screen shot 2016-10-03 at 11 25 50 pm

kouryuu: The identation would look like this is it ok?
screen shot 2016-10-03 at 11 28 28 pm

varjmes: @kouryuu wonderful!

varjmes (approved):

public/index.html
kouryuu: @gr2m what do you think?

gr2m: can you please wrap the `<strong>...</strong>` tags into `<p>...</p>` tags? `strong` is an inline element, it should be wrapped into a block element

kouryuu: @gr2m sure 👍

gr2m (approved):
perfect, thank you for your help and patience with us Rodrigo :)

LowProfileDog:
released as [v24.2.0](https://github.com/hoodiehq/hoodie/releases/tag/v24.2.0) Thanks @kouryuu 👏 

I’ve invited you to our Contributors Team, thanks so much for your contribution :) Are you on twitter by chance? We like giving shout outs to our new contributors, only if you like of course!

kouryuu:
🎉😃  I am so thrilled this community is so good, and friendly.
It feels great to contribute to OSS wait for the next PR's 😉
My twitter is https://twitter.com/scrooge_mc_pato

PR #588, opened 2016-10-02
Re-enable logging

### LINKED ISSUE #572: logging is no longer working
this might be related to https://github.com/hoodiehq/hoodie/pull/478

We are using [npmlog](https://www.npmjs.com/package/npmlog) for logging and to set log levels. And we have a logger plugin that uses Good so that `server.log` calls to get logged using `npmlog` at https://github.com/hoodiehq/hoodie/blob/master/server/plugins/logger.js

But it’s no longer working. All `server.log` calls are ignored :(

gr2m:
closes #572 

Problem was that we registered the `@hoodie/server` plugin before registering the logger plugin, and by default hapi doesn’t do anything when doing `server.log`

### DIFF
--- a/server/index.js
+++ b/server/index.js
@@ -58,19 +58,17 @@ function register (server, options, next) {
   options.PouchDB = PouchDB.defaults(options.db)
   delete options.db
 
-  server.register({
-    register: hoodieServer,
-    options: options
-  }, function (error) {
+  server.ext('onPreResponse', corsHeaders)
+
+  registerPlugins(server, options, function (error) {
     if (error) {
       return next(error)
     }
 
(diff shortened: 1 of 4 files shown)


gr2m:
ping @hoodiehq/maintainers 👀

gr2m (commented):
I’ve tested it locally with hoodie-app-tracker and its integration tests and it looks all good

gr2m:
I’ve created this camp issue: https://github.com/hoodiehq/camp/issues/35 and only remember that logging is not actually working right now, which might be very confusing for people looking into the issue. I hope you don’t mind me merging it in, but please review anyway, I’ll make a follow up PR if needed

Issue #35, opened 2013-04-09
password confirmation screen not used/ignored


xpollen8:
reproduce: in the default app - decline to fill in the 2nd password field, account will be created anyway

gr2m:
that's by design, the Hoodie method doesn't take a confirmation, you need to handle that yourself. But I've added a ticket at https://github.com/hoodiehq/my-first-hoodie/issues/9 to fix it in the example app

gr2m:
thanks for bringing that up!

PR #590, opened 2016-10-02
Raise test coverage to 100%

### LINKED ISSUE #4: 💯 Code Coverage for hoodie
This is a [starter issue](http://hood.ie/blog/starter-issues.html). It’s reserved for participants of [OSCON Contribute](https://ti.to/hoodie/oscon-meet-the-hoodies) until May 20th, 2016. 
### :octocat:  Your First Pull Request?

We would like to take the opportunity to invite someone who did not yet contribute to Hoodie to work on this issue. Follow [@YourFirstPR](https://twitter.com/yourfirstPR) on Twitter for more of these issues across many open source projects :)

JavaScript, Node.js. Knowledge about [npm](https://docs.npmjs.com/) would be helpful but are not required, you’ll figure it out as you go :)
### :dart: The goal

100% code coverage makes sure that every single line of code is covered by tests. And it looks pretty cool once we have all the 💯 for all modules in [Hoodie’s Architecture](https://github.com/hoodiehq/hoodie/tree/master/server#architecture)

The current status is [Coverage Status](https://coveralls.io/r/hoodiehq/hoodie?branch=master)
screen shot 2016-10-01 at 12 30 49
### :clipboard: Step by Step
- [x] claim this issue: assign yourself if you are part of the Hoodie team, otherwise comment below :)
- [x] replace `up for grabs` label with `in progress`
- [x] Bring coverage to 100%. Use the following steps
  - run `npm test`
  - run `node_modules/.bin/nyc report --reporter=lcov`
  - open `coverage/lcov-report/index.html`. It will look something like this: coverage
- [x] If you are all 100%, commit with `test: 100% coverage`. If you have blockers commit with `test: coverage` and move on, we can discuss details in the pull request
- [x] Start a Pull Request. Mention `closes hoodiehq/camp#4` in the description.  
  If this is your first, welcome :tada: :smile: [Here is a great tutorial](https://egghead.io/series/how-to-contribute-to-an-open-source-project-on-github) on how to send a pull request

davelosert:
closes hoodiehq/camp#4

I'm putting the PR already so you can see my progress and to be able to discuss certain steps along the way. :)

### DIFF
--- a/server/plugins/client/index.js
+++ b/server/plugins/client/index.js
@@ -31,28 +31,9 @@ function register (server, options, next) {
               return reject(error)
             }
 
+            writeClientBundle(hasUpdate, options.config.inMemory, bundleTargetPath, bundleBuffer)
+
             resolve(bundleBuffer)
-
-            if (!hasUpdate) {
-              log.info('client', 'bundle is up to date')
-              return
-            }
-
-            if (options.config.inMemory) {
-              log.silly('client', 'running in memory, not writing bundle to ' + bundleTargetPath)
-              return
-            }
-
-            log.info('client', 'bundle is out of date')
-            log.silly('client', 'writing bundle to ' + bundleTargetPath)
-            fs.writeFile(bundleTargetPath, bundleBuffer, function (error) {
-              if (error) {
-                log.warn('client', 'could not write to ' + bundleTargetPath + '. Bundle cannot be cached and will be re-generated on server restart')
-                return
-              }
-
-              log.info('client', 'bundle written to ' + bundleTargetPath)
-            })
           })
         })
       }
@@ -65,3 +46,27 @@ function register (server, options, next) {
 
   next()
 }
+
+function writeClientBundle (hasUpdate, inMemory, bundleTargetPath, bundleBuffer) {
+  if (!hasUpdate) {
+    log.info('client', 'bundle is up to date')
+    return
+  }
+
+  if (inMemory) {
+    log.silly('client', 'running in memory, not writing bundle to ' + bundleTargetPath)
+    return
+  }
+
+  log.info('client', 'bundle is out of date')
+  log.silly('client', 'writing bundle to ' + bundleTargetPath)
+  fs.writeFile(bundleTargetPath, bundleBuffer, function (error) {
+    if (error) {
+      log.warn('client', 'could not write to ' + bundleTargetPath + '. Bundle cannot be cached and will be re-generated on server restart')
+      return
+    }
+
+    log.info('client', 'bundle written to ' + bundleTargetPath)
+  })
+}
+
--- a/test/integration/force-gzip-test.js
+++ b/test/integration/force-gzip-test.js
@@ -73,3 +73,4 @@ function testGzip (group, server, response) {
     server.stop(group.end)
   })
 }
+
--- a/test/integration/handle-404-test.js
+++ b/test/integration/handle-404-test.js
@@ -53,5 +53,26 @@ test('forward all requests that accept html to app', function (group) {
     })
   })
 
+  group.test('send a 404 on a unknown /hoodie-request', function (t) {
+    var server = new Hapi.Server()
+    server.connection({port: 8090})
+    server.register(hapiPluginOptions, function (error) {
+      t.error(error, 'hoodie loads without error')
+
+      server.inject({
+        url: 'http://localhost:8090/hoodie/unkown',
+        headers: {
+          accept: 'text/html'
+        }
+      }, function (response) {
+        t.is(response.statusCode, 404, 'statusCode is 404')
+        t.is(response.result.error, 'Not Found', 'Not Found error')
+
+        server.stop(t.end)
+      })
+    })
+  })
+
   group.end()
 })
+
--- a/test/integration/handle-hoodie-request-test.js
+++ b/test/integration/handle-hoodie-request-test.js
@@ -1,0 +1,71 @@
+var Hapi = require('hapi')
+var test = require('tap').test
+
+var hoodie = require('../../').register
+var hapiPluginOptions = {
+  register: hoodie,
+  options: {
+    inMemory: true,
+    loglevel: 'error'
+  }
+}
+
+require('npmlog').level = 'error'
+
+test('respond to all /hoodie/* - requests with an index.html', function (group) {
+  var server
+
+  group.beforeEach(function (done) {
+    server = new Hapi.Server()
+    server.connection({port: 8090})
+    server.register(hapiPluginOptions, function () {
+      done()
+    })
+  })
+
+  group.afterEach(function (done) {
+    server.stop(done)
+  })
+
+  group.test('respond to /admin/', function (t) {
+    checkRequestForPath(t, '/hoodie/admin/')
+  })
+
+  group.test('repond to /account/', function (t) {
+    checkRequestForPath(t, '/hoodie/account/')
+  })
+
+  group.test('respond to /store/', function (t) {
+    checkRequestForPath(t, '/hoodie/store/')
+  })
+
+  group.test('respond to simple /hoodie/', function (t) {
+    server.inject({
+      url: 'http://localhost:8090/hoodie/unkown',
+      headers: {
+        accept: 'text/html'
+      }
+    }, function (response) {
+      t.is(response.statusCode, 404, 'statusCode is 404')
+
+      server.stop(t.end)
+    })
+  })
+
+  function checkRequestForPath (t, path) {
+    server.inject({
+      url: 'http://localhost:8090' + path + 'unknownSubPath',
+      headers: {
+        accept: 'text/html'
+      }
+    }, function (response) {
+      t.is(response.statusCode, 200, 'statusCode is 200')
+      t.match(response.payload, /<html/, 'response is HTML')
+
+      t.end()
+    })
+  }
+
+  group.end()
+})
+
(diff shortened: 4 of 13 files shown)


gr2m:
looks great so far 👍

server/plugins/client/index.js
```diff
@@ -31,28 +31,24 @@ function register (server, options, next) {
...
+              log.info('client', 'bundle is out of date')
+              log.silly('client', 'writing bundle to ' + bundleTargetPath)
+              fs.writeFile(bundleTargetPath, bundleBuffer, function (error) {
+                if (error) {
+                  log.warn('client', 'could not write to ' + bundleTargetPath + '. Bundle cannot be cached and will be re-generated on server restart')
+                  return
+                }
+
+                log.info('client', 'bundle written to ' + bundleTargetPath)
+              })
```
gr2m: we try to avoid `else` statements, which is whey we add the `return`s, see http://blog.timoxley.com/post/47041269194/avoid-else-return-early

gr2m (changes requested):

davelosert:
@hoodiehq/maintainers : Done with this - test coverage is now at :100: !

I took some notes and questions about some things that caught my eye during implementing the tests that I would like to share and hear your opinions. Sorry in advance for the HUGE amount of text following. :D 

### unit/server/index-test.js

Since I couldn't reach 100% test coverage with integration tests here I created a unit-test-file for the server. As it wasn't in the scope of this pr, this unit-test-file now only contains one test - the error case. It would maybe be nicer if it actually would also cover the whole server.

### unit/logger-test.js & server/plugins/logger.js

I realized quickly that the logger-module itself is not very easy testable. It would be easier if the actual logging-method was exposed from the module or extracted into a submodule.

As I see it, the `logger.js` has 3 responsibilities:
1. Register the plugin as such with the server
2. Create the “GoodCase”-Class as a stream-logger
3. Create the actual log-function for the different events

It would be way nicer (and better testable) to split those 3 responsibilities into three different files. If you agree we could open an issue for that and I would try to create a PR.

### client/index.js
1. I had to refactor the handler in order to make it testable: So far, the promise was resolved BEFORE continuing logic. As the resolved promise is the only way to know from the "outside" when the test i done, I wasn't able to test what `fs` does. Instead of having `return` statements, i replaced the whole thing with an `if-else-cascade`.
2. Something similar to logger: It would be way easier to test if the actual handler-function was exposed or refactored into a different file. Right now I wasn’t able test if `fs` actually writes to the correct path  (or only with lots of code duplication). If the Handler-Function could be created with all those paths as options-object, it would be easier. What do you think?

### handle-404-test.js

When sending a request with `accept: text/html` to the url `/hoodie/unkown`, the server responds with JSON. Maybe there should be a `404.html` page? 

### server/plugins/client/bundle.js

When `browserify.bundle` runs into an error, the error is thrown instead of passed to the callback – effectively killing the server. Shouldn’t this be handled more gracefully by invoking an error-callback instead?

server/plugins/client/index.js
davelosert: I understand. As I wrote in my comment: I was only able to test what was happening after the `resolve`-call by putting the resolve-call behind the rest of the code. I will try to find a better solution here.

davelosert: @gr2m : I used another approach by creating an actual `decideAction`-Method to call. Now the code looks like before - only inside an own method. What do you think?

server/plugins/client/index.js
```diff
@@ -19,6 +19,29 @@ function register (server, options, next) {
...
+    log.silly('client', 'writing bundle to ' + bundleTargetPath)
+    fs.writeFile(bundleTargetPath, bundleBuffer, function (error) {
+      if (error) {
+        log.warn('client', 'could not write to ' + bundleTargetPath + '. Bundle cannot be cached and will be re-generated on server restart')
+        return
+      }
+
+      log.info('client', 'bundle written to ' + bundleTargetPath)
+    })
+  }
```
gr2m: I would suggest we add a `inMemory` and a `bundleTargetPath` argument, so that the function no longer access state outside of its scope. Then move it to the very end of the file, outside of the `register` function. And please define it with `function decideAction (hasUpdate, bundleBuffer) {` instead of `var decideAction = function (hasUpdate, bundleBuffer) {`

server/plugins/client/index.js
```diff
@@ -30,29 +53,9 @@ function register (server, options, next) {
             if (error) {
               return reject(error)
             }
+            decideAction(hasUpdate, bundleBuffer)
```
gr2m: I’m not sure about the naming of the function. Naming is hard :)

What it does is: it checks a few things and eventually writes the bundle to the file system. Maybe let’s call it `writeClientBundle(hasUpdate, inMemory, bundleTargetPath, bundleBuffer)`?

gr2m (changes requested):
wow, this is really, really great work, thanks SO MUCH! I’ve just few minor comments, the rest is great! Wow this made my day :)

server/plugins/client/index.js
davelosert: way better that way - and done!

server/plugins/client/index.js
davelosert: Renamed it.

davelosert:
@gr2m haha 'thanks for the flowers' - as we say in germany. It was my pleasure and I'm happy if you like it. :) 

I made the changes you requested, let me know if there is anything else.

gr2m:
> ### unit/server/index-test.js
> 
> Since I couldn't reach 100% test coverage with integration tests here I created a unit-test-file for the server. As it wasn't in the scope of this pr, this unit-test-file now only contains one test - the error case. It would maybe be nicer if it actually would also cover the whole server.

The error callbacks are often times hard to test. Other projects add a `/* istanbul ignore next */` before the `if (error) { ... }`. I think using a unit test here is a good choice, creating an integration test that makes the registration of `@hoodie/server` fail would be hard to create, understand and maintain

> ### unit/logger-test.js & server/plugins/logger.js
> 
> I realized quickly that the logger-module itself is not very easy testable. It would be easier if the actual logging-method was exposed from the module or extracted into a submodule.
> 
> As I see it, the `logger.js` has 3 responsibilities:
> 1. Register the plugin as such with the server
> 2. Create the “GoodCase”-Class as a stream-logger
> 3. Create the actual log-function for the different events
> 
> It would be way nicer (and better testable) to split those 3 responsibilities into three different files. If you agree we could open an issue for that and I would try to create a PR.

Yes, I agree! It’s a bin of a pain point we have in Hoodie right now, see also https://github.com/hoodiehq/camp/issues/35. I think it would be great if you could create a follow up issue for this, and mention #35, then we can work on it once we have #35 taken care of?

> ### client/index.js
> 1. I had to refactor the handler in order to make it testable: So far, the promise was resolved BEFORE continuing logic. As the resolved promise is the only way to know from the "outside" when the test i done, I wasn't able to test what `fs` does. Instead of having `return` statements, i replaced the whole thing with an `if-else-cascade`.

👍 

> 1. Something similar to logger: It would be way easier to test if the actual handler-function was exposed or refactored into a different file. Right now I wasn’t able test if `fs` actually writes to the correct path  (or only with lots of code duplication). If the Handler-Function could be created with all those paths as options-object, it would be easier. What do you think?

Hm I cannot follow 100%, but I love that you are thinking on how to make the code base better to test. Maybe you could create a follow up issue for it and start a pull request as a base of discussion?

> ### handle-404-test.js
> 
> When sending a request with `accept: text/html` to the url `/hoodie/unkown`, the server responds with JSON. Maybe there should be a `404.html` page? 

Good idea 👍 I’d suggest we create a follow up issue as well, because having 100% test coverage is so awesome to have, I’d like to merge it asap and then iterate, if that’s okay with you?

> ### server/plugins/client/bundle.js
> 
> When `browserify.bundle` runs into an error, the error is thrown instead of passed to the callback – effectively killing the server. Shouldn’t this be handled more gracefully by invoking an error-callback instead?

Good catch, yes! Well, ehm, let’s create a follow up issue as well?

---

Wow, great work, danke David! If you are okay with creating follow up issues, I’d merge your pull request in :) 

Only a few minors, but if you are busy I can take care of them myself :) 
1. Could you make sure that all files end with an empty line? You did it in `test/unit/server/logger-test.js`, but for example not in `test/unit/server/index-test.js`
2. Rebase your PR on top of master, reset your commits and commit
   1. server/\* with `refactor: testability` (or something like that, the `refactor:` prefix is significant
   2. the rest with `test: 100% test coverage`

That’d be 💯

davelosert:
@gr2m : You are very welcome! :smiley: 

So I squashed the commits a little and separated them into one commit for the test-coverage and one for the refactoring. Line-Endings should be correct now too (the reason it was like that: in IntelliJ you actually have to put TWO empty lines at the end of a file in order to make it commit ONE empty line ending to git - that's super weird, I am going to open an Issue at jetbrains).

For all the other tasks I created issue #593 , #594 , #595 and #596 - as I already wrote in `Slack` I won't be able to do much this whole next week so I can not tackle them myself. Feel free to put them into the camp or to let it be grabbed by someone else here.

gr2m (approved):
amazing work, thanks so much!

gr2m:
I’ve invited you to Hoodie’s Contributors team :) Welcome!

And are you on twitter by chance? We like giving new contributors a shoutout – only if you like of course

davelosert:
:) nice, thank you very much.
And of course, my twitter is [@david_losert](https://twitter.com/david_losert) - thanks a lot! :+1:

PR #591, opened 2016-10-03
Rename "bindAddress" option to "address"

capellini:
To align with hapi's server.connection options, rename Hoodie's `bindAddress` command line option to `address`.  Update documentation accordingly and log a deprecation warning.

Closes hoodiehq/hoodie-server#496.

### DIFF
--- a/cli/options.js
+++ b/cli/options.js
@@ -1,5 +1,6 @@
 module.exports = getCliOptions
 
+var log = require('npmlog')
 var pick = require('lodash').pick
 var rc = require('rc')
 var yargs = require('yargs')
@@ -37,10 +38,14 @@ function getCliOptions (projectPath) {
         default: defaults.port,
         describe: 'Port-number to run the Hoodie App on'
       },
+      address: {
+        type: 'string',
+        default: defaults.address,
+        describe: 'Address to which Hoodie binds'
+      },
       bindAddress: {
         type: 'string',
-        default: defaults.bindAddress,
-        describe: 'Address that Hoodie binds to'
+        describe: '[DEPRECATED] Address to which Hoodie binds (see --address)'
       },
       data: {
         type: 'string',
@@ -87,6 +92,11 @@ function getCliOptions (projectPath) {
     .epilogue('Options can also be specified as environment variables (prefixed with "hoodie_") or inside a ".hoodierc" file (json or ini).')
     .wrap(Math.min(150, yargs.terminalWidth()))
     .argv
+
+  if (options.bindAddress) {
+    log.warn('The use of --bindAddress is deprecated. Use the --address option instead.')
+    options.address = options.bindAddress
+  }
 
   // rc & yargs are setting keys we are not interested in, like in-memory or _
   // so we only pick the relevant ones based on they keys of the default options.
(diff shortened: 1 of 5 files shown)


capellini:
Should I use `console.warn` to log the deprecation warning, or is there another logging utility that I should use?

cli/options.js
```diff
@@ -88,6 +92,11 @@ function getCliOptions (projectPath) {
     .wrap(Math.min(150, yargs.terminalWidth()))
     .argv
 
+  if (options.bindAddress) {
+    console.warn('The use of --bindAddress is deprecated. Use the --address option instead.')
```
capellini: Should I use `console.warn` to log the deprecation warning, or is there another logging utility that I should use?

If I should use `console.warn`, any idea how to mock it with `simple-mock` so the output doesn't display when testing?

gr2m: yes, we use `npmlog`.

Like this

``` js
var log = require('npmlog')
// ...
log.warn('The use of --bindAddress is deprecated. Use the --address option instead.')
```

Then in tests you can do 

``` js
require('npmlog').level = 'error'
```

gr2m (approved):
Thanks Rob, looking great! I would like to merge https://github.com/hoodiehq/hoodie/pull/590 first, and then rebase your PR on top of it, as #590 introduces lots of changes to tests

capellini:
> Thanks Rob, looking great! I would like to merge #590 first, and then rebase your PR on top of it, as #590 introduces lots of changes to tests

Ok.  Sounds good. :+1:

varjmes (approved):

varjmes:
lgtm

capellini:
Rebase complete.

In order to get 100% branch coverage, I needed to cover a case where `bindAddress` is converted into `address` and there also needs to be a case where we don't enter the `if` statement, so I had to change the test script to cover both cases.

Since I needed to create an instance of `yargs` with `bindAddress` and one without, I made a `createYargsMock` function.  Also, to avoid code duplication, I made a function to create our `getCliOptions` object for both test cases.

I'm not sure if there's a better way to accomplish what I did, but I'm looking forward to getting feedback!

test/unit/cli/options-test.js
```diff
@@ -2,31 +2,32 @@ var proxyquire = require('proxyquire').noCallThru()
 var simple = require('simple-mock')
 var test = require('tap').test
 
-var yargsApi = {
-  argv: {
-    port: 'cli-port'
-  }
-}
+function createYargsMock (argv) {
+  var yargsApi = {argv}
```
gr2m: we don’t use es6 yet, can you change it to `{argv: arvg}` please?

gr2m (changes requested):
besides the minor comment, this looks great 👍

test/unit/cli/options-test.js
capellini: Certainly.  Change made. Thanks!

gr2m (approved):
great work 👍

gr2m:
released in [v24.1.0](https://github.com/hoodiehq/hoodie/releases/tag/v24.1.0) 👍

PR #592, opened 2016-10-03
Set "start" script to "hoodie" when package is installed with --save or -S

### LINKED ISSUE #2: npm start -- setup

JavaScript, Node.js. Knowledge about `npm` would be helpful but are not required.
### ❓ The Motivation

We try to integrate very tightly with `npm` and its eco system. Instead of creating our own [CLI](https://en.wikipedia.org/wiki/Command-line_interface) tool, we want to use `npm` directly. At this point we only use the `npm start` script to start the Hoodie Server, in future we might want to have all kind of commands like `npm start setup`, `npm start console`, etc. 
### :dart: The goal

When running `npm start setup`, the Hoodie Server should _not_ get started. Instead the following things should happen:
1. Create a `README.md` unless it already exists. Put in dummy content for now
2. Set the "start" script in `package.json` to `"hoodie"` (see [setup instructions](https://github.com/hoodiehq/hoodie#setup))
### :clipboard: Step by Step
- [x] claim this issue (comment below or assign yourself if you can)
- [x] Implement the feature. `hoodie` executes the binary which is [hoodie/bin/start.js](https://github.com/hoodiehq/hoodie/blob/master/bin/start.js) as defined [here](https://github.com/hoodiehq/hoodie/blob/26016aabf8ab17935561264e14d3954dd821e6ba/package.json#L6-L8). Instead of calling the `cli` function, check if the `setup` argument was passed and if yes, do the setup as described :) A first step (and commit) could be a `console.log('setup')`, to get a pull request going. 
- [x] Commit with `feat: hoodie setup`  
  Don’t worry about wording or amount of commits. We will clean up everything before we merge :)
- [x] Start a Pull Request. Mention `closes hoodiehq/camp#2` in the description.  
  If this is your first, welcome :tada: :smile: [Here is a great tutorial](https://egghead.io/series/how-to-contribute-to-an-open-source-project-on-github) on how to send a pull request

### LINKED ISSUE #3: Run setup in `npm install --save hoodie` posthook script
### :octocat:  Your First Pull Request?

We would like to take the opportunity to invite someone who did not yet contribute to Hoodie to work on this issue. Follow [@YourFirstPR](https://twitter.com/yourfirstPR) on Twitter for more of these issues across many open source projects :)

This issue depends on #2.

JavaScript, Node.js, npm post script hooks ([see postinstall](https://docs.npmjs.com/misc/scripts))
### ❓ The Motivation

Currently, setting up hoodie takes 2 steps:
1. `npm install --save hoodie`
2. Set the `"start"` script to `"hoodie"`

We want to get rid of the 2nd step :)
### :dart: The goal

Run the setup from #2 in `"postinstall", but only if the install command is`npm install --save hoodie`or`npm i -S hoodie`, not for e.g.`npm install`
### :clipboard: Step by Step
- [x] claim this issue  
  comment below. Once claimed we add you as contributor to this repository, so we can assign you to the ticket and you can start checking off the checkboxes below as you progress. We will also replace the `up for grabs` label with `in progress`.
- [x] Implement the feature. You’ll need to set `"postinstall"` script to `"npm start setup"` but somehow with an extra check, so that it only runs if the original command was `npm start setup` or `npm install --save hoodie` (or using one of the aliases like `i` for `install`)
- [x] Commit with `feat: run setup on npm install --save hoodie`  
  Don’t worry about wording or amount of commits. We will clean up everything before we merge :)
- [x] Start a Pull Request. Mention `closes hoodiehq/camp#3` in the description.  
  If this is your first, welcome :tada: :smile: [Here is a great tutorial](https://egghead.io/series/how-to-contribute-to-an-open-source-project-on-github) on how to send a pull request

klazich:
**closes hoodiehq/camp#2**
**closes hoodiehq/camp#3**

I have some comments and questions i'll add here tomorrow.

### DIFF
--- a/bin/setup.js
+++ b/bin/setup.js
@@ -1,0 +1,60 @@
+#!/usr/bin/env node
+
+var path = require('path')
+var fs = require('fs')
+var log = require('npmlog')
+
+// use process.env to access npm config environment variables.
+// - https://docs.npmjs.com/misc/config#environment-variables
+// - https://docs.npmjs.com/misc/config#save
+var saveRequested = process.env.npm_config_save
+
+// This block only executes if Hoodie is installed with `-S` or `--save` flags.
+if (saveRequested) {
+  var pathToAppRoot = path.resolve('..', '..')
+  var packageJson = require(path.join(pathToAppRoot, 'package.json'))
+
+  packageJson.scripts = packageJson.scripts || {}
+  packageJson.scripts[ 'start' ] = 'hoodie'
+
+  var newPackageJson = JSON.stringify(packageJson, null, 2)
+
+  fs.writeFile(path.join(pathToAppRoot, 'package.json'), newPackageJson,
+    function (error) {
+      if (error) {
+        log.error('setup', 'Could not create package.json at ' + path.join(pathToAppRoot, 'package.json'))
+        log.error(error)
+        return
+      }
+      log.info('setup', 'Start your Hoodie app with "npm start"')
+    })
+
+  // Create README.md if one is not found.
+  // - https://nodejs.org/api/fs.html#fs_fs_access_path_mode_callback
+  fs.open(path.join(pathToAppRoot, 'README.md'), 'wx',
+    function (error, fd) {
+      if (error) {
+        if (error.code === 'EEXIST') {
+          log.verbose('setup', 'README.md already exists')
+          return
+        } else {
+          log.error('setup', 'Something went wrong when opening README.md at ' + pathToAppRoot)
+          log.error(error)
+          return
+        }
+      }
+
+      // - https://nodejs.org/api/path.html#path_path_parse_path
+      var base = path.parse(pathToAppRoot).base
+      var readMeContents = '# ' + base + '\n' + 'Created with [hoodie](https://github.com/hoodiehq)'
+      fs.writeFile(fd, readMeContents,
+        function (error) {
+          if (error) {
+            log.error('setup', 'Could not create README.md at ' + pathToAppRoot)
+            log.error(error)
+            return
+          }
+          log.info('setup', 'README.md created')
+        })
+    })
+}
--- a/package.json
+++ b/package.json
@@ -65,6 +65,7 @@
     "semantic-release": "semantic-release pre && npm publish && semantic-release post",
     "start": "./bin/start.js",
     "pretest": "standard",
-    "test": "nyc tap --no-cov './test/{unit,integration}/**/*-test.js'"
+    "test": "nyc tap --no-cov './test/{unit,integration}/**/*-test.js'",
+    "postinstall": "node ./bin/setup.js"
   }
 }
--- a/README.md
+++ b/README.md
@@ -60,28 +60,10 @@ npm install --save hoodie
 <!--
   TODO: automate package.json update using postinstall
         https://github.com/hoodiehq/hoodie/issues/477
+        https://github.com/hoodiehq/hoodie/pull/592
 -->
-Add a `"start"` script and set it to `"hoodie"` in your `package.json`. The
-result should look something like this
 
-```json
-{
-  "name": "my-app",
-  "version": "1.0.0",
-  "scripts": {
-    "start": "hoodie",
-    "test": "echo \"Error: no test specified\" && exit 1"
-  },
-  "dependencies": {
-    "hoodie": "^24.0.0"
-  }
-}
-```
-
-There might be more properties and `24.0.0` will likely be a higher number, but
-that’s okay.
-
-Now run `npm start` to start your Hoodie app.
+Now simply use `npm start` to start up your Hoodie app!
 
 You can find a more thorough description in our [Getting Started Guide](http://docs.hood.ie/camp/start/index.html).
 


varjmes:
LGTM as far as I can tell :)

varjmes (approved):

set-start.js
```diff
@@ -0,0 +1,21 @@
+#!/usr/bin/env node
+
+var path = require('path')
+var fs = require('fs')
+
+var save = process.env.npm_config_save
+
+if (save) {
+  var fileName = path.join('..', '..', 'package.json')
```
Haroenv: Wouldn't cwd be better? That way this doesn't need to be updated when files change location.

as suggested by @gr2m in https://github.com/hoodiehq/camp/issues/2#issuecomment-249711353

gr2m: Yes, thanks a lot @Haroenv for reviewing the pull request 👍 Can you please change that @klazich?

set-start.js
```diff
@@ -0,0 +1,21 @@
...
+  file.scripts = file.scripts || {}
+  file.scripts[ 'start' ] = 'hoodie'
+
+  fs.writeFile(fileName, JSON.stringify(file, null, 2),
+    function (err) {
+      if (err) return console.log(err)
+    })
+
+  console.log('Start your Hoodie app with "npm start"' + '\n')
+}
```
gr2m: can we please move this file into `bin/set-start.js`? And how about renaming the script to `bin/setup.js`?

set-start.js
```diff
@@ -0,0 +1,21 @@
...
+
+  file.scripts = file.scripts || {}
+  file.scripts[ 'start' ] = 'hoodie'
+
+  fs.writeFile(fileName, JSON.stringify(file, null, 2),
+    function (err) {
+      if (err) return console.log(err)
+    })
+
+  console.log('Start your Hoodie app with "npm start"' + '\n')
```
gr2m: can you move this line into the `fs.writeFile` callback?

set-start.js
```diff
@@ -0,0 +1,21 @@
...
+  var fileName = path.join('..', '..', 'package.json')
+  var file = require(fileName)
+
+  file.scripts = file.scripts || {}
+  file.scripts[ 'start' ] = 'hoodie'
+
+  fs.writeFile(fileName, JSON.stringify(file, null, 2),
+    function (err) {
+      if (err) return console.log(err)
+    })
```
gr2m: two minor things:
1. We try to avoid abbreviations in Hoodie’s code for clarity
2. Can you wrap `return console.log(err)` into `{}` with line breaks?

set-start.js
```diff
@@ -0,0 +1,21 @@
+#!/usr/bin/env node
+
+var path = require('path')
+var fs = require('fs')
+
+var save = process.env.npm_config_save
```
gr2m: is `npm_config_save` documented somewhere? If yes, could you add a comment above with the URL to it? If not, you could mention that `process.env.npm_config_save` is true when running `npm install` with either `-S` or `--save`.

Maybe what would also help is to rename the variable to something more expressive, like `hasNpmSaveFlag`?

package.json
```diff
@@ -64,6 +64,7 @@
     "semantic-release": "semantic-release pre && npm publish && semantic-release post",
     "start": "./bin/start.js",
     "pretest": "standard",
-    "test": "nyc tap --no-cov './test/{unit,integration}/**/*-test.js'"
+    "test": "nyc tap --no-cov './test/{unit,integration}/**/*-test.js'",
+    "postinstall": "node ./set-start.js"
```
gr2m: I _think_ that `postinstall` is run everytime someone runs `npm install` in the app, not only if the `hoodie` package was installed. Could you please test it by installing another package and see if `node ./set-start.js` is run? If it is, we should somehow check if the `hoodie` package was installed, if not, don’t run the code.

gr2m (changes requested):
I think this is a great start, short and elegant, thanks! I left a few comments, I hope they all make sense? Please let me know if you have any questions

klazich (commented):
- Added links to documentation for npm config save environment variable
- Refactored variable names to remove abbreviations and making them more expressive.
- Moved console.log into fs.writeFile cb and added braces to if block.
- Moved JSON.stringify to variable for easier reading.

set-start.js
```diff
@@ -0,0 +1,29 @@
+#!/usr/bin/env node
+
+var path = require('path')
+var fs = require('fs')
+
+// use process.env to access npm config environment variables
+var saveRequested = process.env.npm_config_save
+// https://docs.npmjs.com/misc/config#environment-variables
+// https://docs.npmjs.com/misc/config#save
```
gr2m: can you move these two comments above `var saveRequested = process.env.npm_config_save` please?

gr2m (changes requested):

gr2m:
there are still a few things open from the reviews, are you still looking into these? Sorry don’t mean to push just want to make sure you don’t miss those :)

set-start.js
klazich: No problem!

And yes, I'm still chugging along on your reviews and comments (thank you guys by the way!). I meant to reply to each last night but ended up stumbling across some interesting npm script stuff I'd never read before and fell down a rabbit hole of documentation and stackoverflow.

gr2m: glad you learn new interesting things along the way :) that’s the way it should be

set-start.js
klazich: When I tried to implement this I had some trouble. `process.cwd()` would always point to the hoodie package in the node_modules and not the project dir. I believe npm executes package.json scripts in a new shell or cmd if that script is called by something else . Below is some output from some experimenting I was doing. 

```
/**
 * The package.json for a-package contains:
 *   "script": { "postinstall": "node ./bin/makefile.js" }
 * a-package installed to app from command line:
 *    c:dev\sandbox\app> npm install ../a-package
 * Below is what is logged during install of a-package to app
 */
process.cwd()                     // c:\dev\sandbox\app\node_modules\a-package

var pathArray1 = [process.cwd(), 'package.json']
path.join(...pathArray1)          // c:\dev\sandbox\app\node_modules\a-package\package.json
path.resolve(...pathArray1)       // c:\dev\sandbox\app\node_modules\a-package\package.json

var pathArray2 = ['..', '..', 'package.json']
path.join(...pathArray2)          // ..\..\package.json
path.resolve(...pathArray2)       // c:\dev\sandbox\app\package.json
```

set-start.js
klazich: I changed it to `saveRequested`. when looking at npm code I noticed that's what they use.

```
// npm/lib/install.js
var saveRequested = require('./install/save.js').saveRequested
```

set-start.js
klazich: 👍

set-start.js
klazich: 👍

set-start.js
klazich: I just pushed this

package.json
klazich: I've been experimenting with this by installing packages in different ways and everything seems to be working the way its supposed to. Though of course that doesn't mean it won't ever break. So I'd like to add a check like you said.

set-start.js
gr2m: great call on looking it up in npm code 👍 Thanks for the efforts!

package.json
gr2m: I think you are right, the`postinstall` script is run _only_ after the package is installed that defines the `postinstall` script. I’ve run a few tests myself and it seems like it. Makes sense anyway :) Thanks for checking!

gr2m: jup, confirmed https://twitter.com/othiym23/status/784107551340232705 👍

klazich:
I added the README check and creation (Part of hoodiehq/camp#2).

Thoughts?

bin/setup.js
```diff
@@ -0,0 +1,53 @@
...
+
+  packageJson.scripts = packageJson.scripts || {}
+  packageJson.scripts[ 'start' ] = 'hoodie'
+
+  var newPackageJson = JSON.stringify(packageJson, null, 2)
+
+  fs.writeFile(path.join(pathToAppRoot, 'package.json'), newPackageJson,
+    function (error) {
+      if (error) {
+        console.error(error)
```
gr2m: can you please use `npmlog` like we do in other parts of `hoodie`? And maybe add a bit of an explanation, too?

``` js
var log = require('npmlog')
log.error('setup', 'Could not create package.json at ' + path.join(pathToAppRoot, 'package.json')
log.error(error)
```

bin/setup.js
```diff
@@ -0,0 +1,53 @@
...
+        return
+      }
+      console.log('Start your Hoodie app with "npm start"\n')
```
gr2m: use `log.info('setup', 'Start your Hoodie app with "npm start"\n')`

bin/setup.js
```diff
@@ -0,0 +1,53 @@
...
+    })
+
+  // Create README.md if one is not found.
+  // - https://nodejs.org/api/fs.html#fs_fs_access_path_mode_callback
+  fs.open(path.join(pathToAppRoot, 'README.md'), 'wx',
+    function (error, fd) {
+      if (error) {
+        if (error.code === 'EEXIST') {
+          // console.error('README.md found')
```
gr2m: `log.verbose('setup', 'README.md already exists')`

bin/setup.js
```diff
@@ -0,0 +1,53 @@
...
+          return
+        } else {
+          throw error
+        }
+      }
+
+      fs.writeFile(fd, '# HOODIE!',
+        function (error) {
+          if (error) {
+            console.error(error)
```
gr2m: like above, add a bit context that README could not be written and then log error itself

bin/setup.js
```diff
@@ -0,0 +1,53 @@
...
+            return
+          }
+          console.log('README.md created.\n')
```
gr2m: See above

gr2m (changes requested):
okay this is great, I tested it with `npm i -S https://github.com/klazich/hoodie.git#npm-start-setup` and it works 👍 

Just a few minor notes regarding logging. 

And also please update [hoodie’s README’s setup instructions](https://github.com/hoodiehq/hoodie/blob/35072299c979a8f33ac8fafaca513880a0f883b4/README.md#setup) :)

klazich:
@gr2m I'll be working on your changes this weekend. 👍

bin/setup.js
```diff
@@ -0,0 +1,53 @@
...
+  // Create README.md if one is not found.
+  // - https://nodejs.org/api/fs.html#fs_fs_access_path_mode_callback
+  fs.open(path.join(pathToAppRoot, 'README.md'), 'wx',
+    function (error, fd) {
+      if (error) {
+        if (error.code === 'EEXIST') {
+          // console.error('README.md found')
+          return
+        } else {
+          throw error
```
klazich: @gr2m is it correct to throw this error here or should it be logged like above with package.json?

gr2m:
This looks good for me, are you done from your side @klazich?

klazich:
Yep. You guys got it from here? Thanks for the opportunity to help out!

gr2m (approved):

PR #598, opened 2016-10-05
Updating Hoodie for Latest `good`

### LINKED ISSUE #35: 🎃🕵💻 Update hoodie to latest good

mbad0la:
This PR closes hoodiehq/camp#35

Progress : 
- [x] Clone
- [x] Test
- [x] Update `good`
- [x] Test (and find the test/logging to fail)
- [x] Translate old endpoints into newer `good` endpoints making necessary changes
- [x] Create/Run unit-tests and test logging
- [x] Wrap-up


gr2m (approved):
This looks great and I tested it locally, all good! I’m just wondering if there is anything we should do about https://github.com/hoodiehq/hoodie/pull/600/files? I think @capellini mentioned that we can remove it once we upgrade to good@7?

capellini:
> I’m just wondering if there is anything we should do about https://github.com/hoodiehq/hoodie/pull/600/files? I think @capellini mentioned that we can remove it once we upgrade to good@7?

On the latest version of Good, they [guard against calling `getLog`](https://github.com/hapijs/good/blob/b8f923eee4e90230cf980a2e5d883a20e3b26738/lib/utils.js#L83) and make sure that `route.settings.log === true` before calling `getLog`.  So we can safely remove the line added in PR #600 and we will no longer get the `Error: Request logging is disabled` error.

My question is (and, pardon once again for the Hoodie newbie question) -- do we want to turn on Good logging all the time?  If so, then we're safe leaving that line in `cli/hapi-options.js`.

Alternatively, is this something that we would want to allow a user to configure (i.e. allow them to turn logging on and off using a hoodie command line switch)?  If so, then maybe we want to put in some logic to provide a default value that the user can override.

gr2m:
> My question is (and, pardon once again for the Hoodie newbie question) -- do we want to turn on Good logging all the time? If so, then we're safe leaving that line in cli/hapi-options.js.

Yes, we want to turn on logging at all time, it is configurable with `--loglevel`, and it can be turned off with `--loglevel=silent`

So this is good to merge then?

capellini:
> So this is good to merge then?

Well, if we're talking about the line in question on #600, then yes, a rebase is all that we need to put that line in.

But I think that there may be an issue elsewhere in the code, specifically in [these lines](https://github.com/mbad0la/hoodie/blob/c4c555ce1b7fa642e8360d4373a3997ea1738177/server/plugins/logger.js#L31-L33), I don't think that the callback is ever invoked.  I think that if we want to do something like what's in this callback -- specifically to filter messages and log appropriately -- we have to create our own stream handler to do this?

I created a test instance of Hoodie and `npm link`ed to this branch. Right now, nothing is logged. :frowning_face:

mbad0la:
Hi @capellini 

> But I think that there may be an issue elsewhere in the code, specifically in these lines.

I had my concerns for the same. I wasn't able to find something related to custom logging callbacks in the docs, so I kind of re-used what was already there in the code.

Can you explain to me how I can locally test out whether there are some logs generated or not? Maybe I might figure out what to do.

capellini:
Hi @mbad0la.  Sure thing -- I'm more than happy to help out.

To set up hoodie for a manual test, I followed steps similar to what's in [this issue](https://github.com/hoodiehq/hoodie-store-client/issues/95), namely:
- In an empty directory, `npm init -y` and hit enter until done.
- Install hoodie: `npm i hoodie -S`.
- Open package.json and add this into the root object:

``` js
"scripts": {
  "start": "hoodie"
}
```
- Add a `public/index.html` file that has `<script src="/hoodie/client.js"></script>` in it.
- Start the app with `npm start -- --in-memory`
- Navigate in a browser to `http://localhost:8080/hoodie/admin/` 

If logging is working, something like the following lines should show up on your terminal:

```
http response 2016-10-09T04:50:44.517Z - 127.0.0.1 - GET /hoodie/admin/ 200 34ms
http response 2016-10-09T04:50:44.575Z - 127.0.0.1 - GET /hoodie/admin/assets/vendor-d41d8cd98f00b204e9800998ecf8427e.css 200 43ms
http response 2016-10-09T04:50:44.586Z - 127.0.0.1 - GET /hoodie/admin/assets/admin-c2f30fd7d34215bdb578a397c7baf297.css 200 43ms
http response 2016-10-09T04:50:44.596Z - 127.0.0.1 - GET /hoodie/admin/assets/admin-f542b3d2135107bff7b691b14dd20872.js 200 40ms
http response 2016-10-09T04:50:44.592Z - 127.0.0.1 - GET /hoodie/admin/assets/vendor-8d2ede4c35a4ecbf8cf54e80aa74b030.js 200 119ms
http response 2016-10-09T04:50:45.365Z - 127.0.0.1 - GET /hoodie/account/api/accounts 200 21ms
```

If you want to test different log levels, you can also use `--loglevel` (e.g. `npm start -- --inMemory --loglevel=silly`.

As far as what we need to do -- We need to create a stream transformer that does what the `logger` function used to do. I took a template from `good-console` and came up with [this gist](https://gist.github.com/capellini/d633428b489d2160323787ef0693ebdf).

This file is still in ES6 format, so we'll need to convert it to use ES5 syntax.  I put the contents of this file into the `server/plugins/` directory and called it `good-hoodie.js`.

The next thing that we need to do is delete [these lines](https://github.com/mbad0la/hoodie/blob/c4c555ce1b7fa642e8360d4373a3997ea1738177/server/plugins/logger.js#L31-L33).

Then, I put another object in [the hoodieReporter](https://github.com/mbad0la/hoodie/blob/5a455d1aef8f8a2ef761f63eb2429bc8350c6872/server/plugins/logger.js#L18-L36) like so:

``` js
        hoodieReporter: [
          {
             // keep the good-squeeze object the same as before, 
             // other than removing the aforementioned lines
          }, {
            module: path.join(__dirname, 'good-hoodie')
          }
        ]
```

(note that we're using `path` now, so be sure to `require` it at the top of the `logger.js` file).

Now, if you restart hoodie, you should see some log lines showing up when you hit `http://localhost:8080/hoodie/admin/`.  You should see even more show up if you run hoodie with `--loglevel=silly` as discussed above.

Hope this helps!  I'm more than willing to work with you on this, if you'd like.

mbad0la:
Thanks @capellini 

I'm trying to resolve this myself, as I want to learn how this works. But if I'm unable to figure it out, i'll add you to my fork so that you may build up on top of my commits. Thanks for your assistance :)

capellini:
> Thanks @capellini

You're most welcome.

> I'm trying to resolve this myself, as I want to learn how this works. But if I'm unable to figure it out, i'll add you to my fork so that you may build up on top of my commits.

Sounds great.  I also learn best by doing.  Please let me know if I can help in any way.

mbad0la:
@capellini @gr2m logging works now. 

But due to the addition of `logger` logic in form of a transformation in the good pipeline, all the existing tests for logging are redundant. How do we work around that?

capellini:
@mbad0la -- great work!

The first thing that I noticed is (at least for now) we need to add `'use strict'` to the to of the script, or it will complain about block-scoped declarations.

As far as the tests are concerned, for the unit tests, we already have an `npmlogMock` variable in `test/unit/server/logger-test.js`, so we need to replace `logger` in that file with an instance of a `HoodieTransform` and call that class' `_transform` method, rather than calling `logger` in the code.

As far as integration tests -- maybe we want to test that log entries are being sent by mocking out `npmlog` again and `request`ing something to ensure that we are logging?

Another thing we need to do -- since I used existing `good` transforms as a template for `HoodieTransform`, the files are in es6 format and we don't use es6 in Hoodie yet.  So we would need to translate the es6 to es5.  That would probably obviate the need for the `'use strict'` line, as well.

gr2m:
Just want to say thanks you two for tackling this issue, I learn a lot and really appreciate you collaborating on this 👍

mbad0la:
@capellini i'll refactor the code to es5. Can you work on writing the tests on top of my commits? I'll add you to my fork once I'm done with the refactoring.

server/plugins/good-hoodie.js
```diff
@@ -0,0 +1,63 @@
...
+      log[level](
+        data.event,
+        new Date(data.timestamp).toISOString(),
+        data.tags.length ? data.tags : data.data,
+        data.tags.length ? data.data : ''
+      )
+    }
+
+    log.silly(data.event, data)
+    next(null)
```
capellini: @mbad0la - one thing I just noticed -- just in case in the future we decide to chain more transforms after `HoodieTransform`, we should send `data` back with the callback, so this should be `next(null, data)`

mbad0la: @capellini I suppose as we are using `npmlog` to do our logging, it won't make sense to send any data with the callback as the custom stream transform would always be the last one. Am I wrong?

capellini: Right now that's correct.  There's nothing that's chained after `good-hoodie`.  But passing the data through the callback doesn't hurt and if we should decide to chain something after the `good-hoodie` transform in the future, for example:

``` js
        hoodieReporter: [
          {
            module: 'good-squeeze',
            name: 'Squeeze',
            args: [
              {
                ops: '*',
                response: '*',
                log: '*',
                error: '*',
                request: '*'
              }
            ]
          },
          {
            module: '../../../server/plugins/good-hoodie'
          },
          {
            module: 'good-squeeze',
            name: 'SafeJson'
          },
          {
            module: 'good-file',
            args: ['./mylog.out']
          }
        ]
```

then nothing would be written to `mylog.out` because the messages are not passed through `good-hoodie`.

mbad0la: @capellini , I've sent you an invite to my fork.

capellini:
@gr2m - ready for review.

gr2m:
thanks, looking into it! 

By the way, for testing `hoodie`, you don’t need to create a new app and link it in, running `npm start` in the `hoodie` repository does the same thing

server/plugins/logger.js
```diff
@@ -3,88 +3,33 @@ module.exports.register.attributes = {
...
+                ops: '*',
+                response: '*',
+                log: '*',
+                error: '*',
+                request: '*'
+              }
+            ]
+          },
+          {
+            module: '../../../server/plugins/good-hoodie'
```
gr2m: can you explain this path? Wouldn’t `'./good-hoodie'` be enough here?

gr2m (changes requested):
just a minor questions, great work you two!

server/plugins/logger.js
capellini: Unfortunately, the file isn't being `require`d here.  Instead, that's being done somewhere inside of good.  So the path needs to be relative to where the good file is located.

I recommend that we change it to either `path.join(__dirname, 'good-hoodie')` or `path,join(process.cwd(), 'good-hoodie')`, in case the folks at hapi decide to change the location of the file in question. 

Another option is to create a new repository and then we could reference it without a relative path, like `module: '@hoodie/good-hoodie'`, or something like that.

mbad0la: I agree with @capellini that using `path` will be a lot more cleaner. I had assumed that we might not change the location of the file and hence, didn't want to require additional modules.

Btw, thanks for writing the tests @capellini :+1:

gr2m: okay let’s go with `path.join(__dirname, 'good-hoodie')`

mbad0la: Done @gr2m

capellini: > Btw, thanks for writing the tests @capellini

You're most welcome @mbad0la

gr2m (approved):

gr2m:
Great work @mbad0la & @capellini 👏 Your work was released as [v24.2.4](https://github.com/hoodiehq/hoodie/releases/tag/v24.2.4)

@mbad0la I’ve invited you to Hoodie’s Contributors Team, welcome :wave:🐶 

Are you on twitter by chance? We like giving shoutouts to our new contributors, only if you like of course :)

mbad0la:
Thanks @gr2m 

And sure! My twittter handle is `@mbad0la` as well.

PR #600, opened 2016-10-05
Enable request logging to prevent error

### LINKED ISSUE #599: Error: Request logging is disabled
During some of the Travis CI builds, like [this one](https://travis-ci.org/hoodiehq/hoodie-app-tracker/jobs/164834321), we receive the following error message (on line 2592 of the log output, in this case):
screenshot from 2016-10-05 09-10-39

This error is caused by a [recent PR from hapi](https://github.com/hapijs/hapi/issues/3306) that [requires a route's `log` property](https://github.com/hapijs/hapi/blob/ed5e75e467e7c5744f6dc58cd067ae1d8bab18a5/lib/request.js#L252) to be `true` in order to use the `getLog` method (which appears to be `false` by default).  However, in the version of Good that Hoodie is currently using, the response event payload in `lib/utils.js` has a `log` property that calls `getLog` without checking if the route's `log` property is set to `true`. 

Note that in the most recent version of Good [they have fixed this issue](https://github.com/hapijs/good/blob/b8f923eee4e90230cf980a2e5d883a20e3b26738/lib/utils.js#L83) by checking the `log` property before calling `getLog`, so the long-term fix would be to use the most recent version of Good, for which we already have a [pending issue](https://github.com/hoodiehq/camp/issues/35).  

In the near term, however, as a temporary patch to prevent the error from happening, we could add a `log` property to [hapi-options.js](https://github.com/hoodiehq/hoodie/blob/a80f13b6841539e0a57378d87bd5bf1c5f3a1f3c/cli/hapi-options.js#L6-L12) so the code looks similar to this:

``` js
  var hapiOptions = {
    server: {},
    connection: {
      port: options.port,
      address: options.address,

      // logging added explictly here
      routes: { log: true },    
    }
  }
```

Fixing this won't make the Travis build pass, as there are other errors, but making a temporary patch to this will at least get rid of this error in the test.

capellini:
Enable request logging in hapi to prevent error during Travis CI builds.

Closes #599.

### DIFF
--- a/cli/hapi-options.js
+++ b/cli/hapi-options.js
@@ -7,7 +7,8 @@ function getHapiOptions (options) {
     server: {},
     connection: {
       port: options.port,
-      address: options.address
+      address: options.address,
+      routes: { log: true }
     }
   }
 


capellini:
Since this is a temporary fix until we can upgrade Good, is it a good idea to somehow mark this so that we know to change it once we upgrade?  I could add a `TODO` comment, but I know that people have very strong opinions about whether or not they should be used.

capellini:
On second thought, maybe it's not a temporary fix, if we want logging enabled.  The default option is `false`.  So if we want logging, then this would have to be included anyway.

gr2m (approved):

gr2m:
released in https://github.com/hoodiehq/hoodie/releases/tag/v24.2.1

Issue #594, opened 2016-10-03
Refactor client/index.js for better testability


davelosert:
Taken from #590 - @gr2m asked me to open another issue for it. 

Regarding [client/index.js](https://github.com/hoodiehq/hoodie/blob/master/server/plugins/client/index.js)

> Something similar to logger (see #593): It would be way easier to test if the actual handler-function was exposed or refactored into a different file. Right now I wasn’t able test if fs actually writes to the correct path (or only with lots of code duplication). If the Handler-Function could be created with all those paths as options-object, it would be easier. What do you think?

With this I meant: You have to jump through a little hoop to get to test the function passed to the `handler`-property given to the `server.router()`-function - as you have to mock the `server` object and then get the handler from the argument passed to it - simplified it looks like this:

```
var simple = require('simple-mock')
var client = require('server/plugins/client');
var mockServer = {
      route: simple.stub()
}
var next = simple.stub()
client.register(mockServer, {config: {}}, next)
// All this just to get to the handler-function I want to test
var handler = mockServer.route.lastCall.arg[0].handler
```

If the actual function passed to `handler` would be extracted into an own file, it would be better testable.
As it depends on some paths (`hoodieClientModulePath`, `hoodieClientPath` and `bundleTargetPath`) I think a factory would be nice, so it could look something like:

```
var createHandler = require('./handler-factory.js)
var hoodieClientModulePath = path.dirname(require.resolve('@hoodie/client/package.json'))
var hoodieClientPath = path.join(hoodieClientModulePath, 'index.js')
var bundleTargetPath = path.join(options.config.data || '.hoodie', 'client.js')

server.route([{
    method: 'GET',
    path: '/hoodie/client.js',
    handler: createHandler(hoodieClientModulePath, hoodieClientPath, bundleTargetPath)
  }])
```

The upside: It would also be testable if the `fs.writeFile()` call actually writes to the `bundleTargetPath` as with the factory one would have directly control over that from the outside (right now thats not possible to be passed in).

davelosert:
I'm going to work on this one...

gr2m:
done via https://github.com/hoodiehq/hoodie/issues/606 👍

PR #606, opened 2016-10-15
Refactor client/index.js for better testability

davelosert:
PR for #594 .

As stated I extracted a factory functions and added several more tests for the paths given to the `bundleClient` and `fs` functions.

I also added a first test for the paths that get built within the `client/index.js`. I realized quickly that thre is made use of `require.resolve` - which is not mockable and therefore the built path is not testable right now.

`require.resolve` is also used within the `server/public.js` file. I think it could be a good idea to take all those `require.resolve` calls into an own separate module like `modulePathResolver` that could look like this:

```
var path = require('path')
var modulePathResolver = {
   getClientPath() : function () {
      return require.resolve('@hoodie/client/package.json')
  }
  getPublicPath : function () {
    return path.join(require.resolve('../../package.json'), '..', 'public')
  }
  getAccountPublicPath : function () {
    return path.join(require.resolve('@hoodie/account/package.json'), '..', 'public')
  }
  // ... and so on for all modulePaths
}
```

@gr2m What do you think about that? Should I raise another issue for it?

### DIFF
--- a/server/plugins/client/bundle-handler-factory.js
+++ b/server/plugins/client/bundle-handler-factory.js
@@ -1,0 +1,32 @@
+module.exports = createBundleHandler
+
+var log = require('npmlog')
+
+var bundleClient = require('./bundle')
+var writeClientBundle = require('./bundle-writer')
+
+function createBundleHandler (hoodieClientPath, bundleTargetPath, bundleConfig) {
+  var bundlePromise
+
+  return function (request, reply) {
+    if (!bundlePromise) {
+      bundlePromise = new Promise(function (resolve, reject) {
+        log.silly('client', 'bundling')
+        bundleClient(hoodieClientPath, bundleTargetPath, bundleConfig, function (error, bundleBuffer, hasUpdate) {
+          if (error) {
+            return reject(error)
+          }
+
+          writeClientBundle(hasUpdate, bundleConfig.inMemory, bundleTargetPath, bundleBuffer)
+
+          resolve(bundleBuffer)
+        })
+      })
+    }
+
+    bundlePromise.then(function (buffer) {
+      reply(buffer).bytes(buffer.length).type('application/javascript')
+    }).catch(reply)
+  }
+}
+
--- a/test/unit/client/client-test.js
+++ b/test/unit/client/client-test.js
@@ -7,29 +7,17 @@ require('npmlog').level = 'error'
 test('client', function (group) {
   var mockServer
   var client
-  var bundleStub
-  var fsStub
-  var npmlogStub
+  var createBundleHandlerStub
 
   group.beforeEach(function (done) {
-    npmlogStub = {
-      info: simple.stub(),
-      silly: simple.stub(),
-      warn: simple.stub()
-    }
+    createBundleHandlerStub = simple.stub()
 
-    bundleStub = simple.stub()
-    fsStub = {
-      writeFile: simple.stub()
-    }
     mockServer = {
       route: simple.stub()
     }
 
     client = proxyquire('../../../server/plugins/client/index', {
-      './bundle': bundleStub,
-      'fs': fsStub,
-      'npmlog': npmlogStub
+      './bundle-handler-factory': createBundleHandlerStub
     })
 
     done()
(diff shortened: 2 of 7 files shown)


gr2m:
> I think it could be a good idea to take all those require.resolve calls into an own separate module like modulePathResolver

Good idea 👍

server/plugins/client/bundle-handler-factory.js
```diff
@@ -0,0 +1,54 @@
+module.exports.createHandler = createHandler
```
gr2m: why not simply `module.exports = createHandler`? That’s what we usually do: 1 file - 1 method :)

server/plugins/client/bundle-handler-factory.js
```diff
@@ -0,0 +1,54 @@
...
+          resolve(bundleBuffer)
+        })
+      })
+    }
+
+    bundlePromise.then(function (buffer) {
+      reply(buffer).bytes(buffer.length).type('application/javascript')
+    }).catch(reply)
+  }
+}
```
gr2m, line 31: We usually don’t do factories, because they are harder to test. We usually prefer pre-binded methods (but that doesn’t mean that it’s always the right way, this is just a suggestion, up for discussion :)

``` js
module.exports = createHandler
var bundlePromise

function createHandler (hoodieClientPath, bundleTargetPath, bundleConfig, request, reply) {
  if (!bundlePromise) {
    bundlePromise = new Promise(function (resolve, reject) {
      log.silly('client', 'bundling')
      bundleClient(hoodieClientPath, bundleTargetPath, bundleConfig, function (error, bundleBuffer, hasUpdate) {
        if (error) {
          return reject(error)
        }

        writeClientBundle(hasUpdate, bundleConfig.inMemory, bundleTargetPath, bundleBuffer)

        resolve(bundleBuffer)
      })
    })
  }

  bundlePromise.then(function (buffer) {
    reply(buffer).bytes(buffer.length).type('application/javascript')
  }).catch(reply)
}
```

And then in `server/plugins/client/index.js` we could do 

``` js
  server.route([{
    method: 'GET',
    path: '/hoodie/client.js',
    handler: createBundleHandler.bind(null, hoodieClientPath, bundleTargetPath, options.config)
  }])
```

server/plugins/client/bundle-handler-factory.js
```diff
@@ -0,0 +1,54 @@
...
+  log.silly('client', 'writing bundle to ' + bundleTargetPath)
+  fs.writeFile(bundleTargetPath, bundleBuffer, function (error) {
+    if (error) {
+      log.warn('client', 'could not write to ' + bundleTargetPath + '. Bundle cannot be cached and will be re-generated on server restart')
+      return
+    }
+
+    log.info('client', 'bundle written to ' + bundleTargetPath)
+  })
+}
```
gr2m: we could also move this into its own file or exporting at `module.exports.internals.writeClientBundle` for unit testing?

gr2m (changes requested):
This is a great pull request, thanks so much for your great work! I’ve some feedback on how we could maybe simplify implementation and testing, up for discussion :)

server/plugins/client/bundle-handler-factory.js
davelosert: I usually prefer factories for the reason as the word 'factory' itself already makes it pretty clear what they are doing. :)

What weighs even stronger is consistency with the rest of the code. In this particular case we have to consider one thing though: The `bundlePromise` - variable in your example is kind of "global" within the module. I don't assume it is a use case right now, but if you would `require` the createHandler twice from different clients, the second handler would already find the `bundlePromise` not being undefined and therefore never call `bundleClient` (with a maybe different configuration than the first). 

You still think we should do it like this? Or maybe do it with `bind` but still wrap the "createHandler" into a function for a private scope?

server/plugins/client/bundle-handler-factory.js
davelosert: Haha, I was actually thinking about doing this but then thought it is not that complex to test right now.
But of course, it would make things even easier! Shall I do it within this PR or open another issue after we are done here?

server/plugins/client/bundle-handler-factory.js
davelosert: Ah - i actually oriented myself in how it was done within the `index.js`:

```
module.exports.register = register
module.exports.register.attributes = {
  name: 'hoodie-local-client',
  dependencies: 'inert'
}
```

But maybe it's different there because it's the index and you add some more attributes to it?
Easy change I can make, though :)

server/plugins/client/bundle-handler-factory.js
gr2m: you are right, and I think we need to bundle more than just `hoodie.js` soon. We could probably combine the two, have a factory that does nothing but defining the bundle variable and then pass it into a pre-bound function like my suggestion above? So the factory method would be only two lines, and we can test the pre-bound function quite easily if we want to

server/plugins/client/bundle-handler-factory.js
gr2m: `index.js` is because it’s a hapi plugin, which has certain requirement: http://hapijs.com/tutorials/plugins

server/plugins/client/bundle-handler-factory.js
gr2m: I think you can make this part of this PR

server/plugins/client/bundle-handler-factory.js
davelosert: You mean something like this?

```
function createHandler(hoodieClientPath, bundleTargetPath, bundleConfig){
   var bundlePromise
   return bundleHandler.bind(null, hoodieClientPath, bundleTargetPath, bundleConfig, bundlePromise)
}

function createHandler (hoodieClientPath, bundleTargetPath, bundleConfig, bundlePromise, request, reply) {
  if (!bundlePromise) {
      ...
  }
```

In that case the `bundlePromise` within the factory-function won't every be changed as it isn't passed by reference, is it? Or did you mean it differently?

test/unit/client/client-test.js
```diff
@@ -4,32 +4,20 @@ var test = require('tap').test
 
 require('npmlog').level = 'error'
 
-test('client', function (group) {
+test('index', function (group) {
```
gr2m: why did you change it from client to index?

gr2m (changes requested):

davelosert:
@gr2m I squashed the commits now. As soon as your saucelabs tests are working again I'd say this is ready to be merged! :)

gr2m:
Thanks 👍 I still didn’t figure out the selenium / saucelabs problems but I have it on my list

gr2m:
just a short update, I did not forget about the PR, we still did not resolve the selenium tests, sorry it takes so long to finish up this PR

davelosert:
No problem. Just merge it as soon as it's ready ;)

gr2m (approved):
all right, I could finally fix the tests of the tracker app and it all looks good now. I rebased your PR on the latest `master` branch and will merge it once CI is green. Thanks so much for your great work 🙌

davelosert:
:hand: great!

gr2m:
✋ 💯

PR #607, opened 2016-10-16
fix: Serve an HTML 404 when Accept: text/html is set

### LINKED ISSUE #595: Create 404.html page for accept: text/html requests
Taken from #590 - @gr2m  asked me to open another issue for it.

> When sending a request with accept: text/html to the url /hoodie/unkown, the server responds with JSON. Maybe there should be a 404.html page?

I guess thats rather clear :smile:

jamietanna:
closes #595.

Also requires a change to [Hoodie-Store](https://github.com/hoodiehq/hoodie-store), which is at https://github.com/hoodiehq/hoodie-store/pull/35.

### DIFF
--- a/server/plugins/public.js
+++ b/server/plugins/public.js
@@ -94,15 +94,15 @@ function register (server, options, next) {
     }
 
     var is404 = response.output.statusCode === 404
-    var isHTML = /text\/html/.test(request.headers.accept)
+    var isResource = /(text\/css|application\/json|application\/javascript|image\/png|image\/jpeg)/.test(request.headers.accept) || /.(css|js|png|jpg)$/.test(request.path)
     var isAccountPath = /^\/hoodie\/account\//.test(request.path)
     var isAdminPath = /^\/hoodie\/admin\//.test(request.path)
     var isStorePath = /^\/hoodie\/store\//.test(request.path)
     var isHoodiePath = /^\/hoodie\//.test(request.path)
 
     // We only care about 404 for html requests...
-    if (!is404 || !isHTML) {
-      return reply.continue()
+    if (is404 && isResource) {
+      return reply(fs.createReadStream(path.join(hoodiePublicPath, '404.html'))).code(404)
     }
 
     if (isAccountPath) {
@@ -115,7 +115,7 @@ function register (server, options, next) {
       return reply(fs.createReadStream(path.join(storePublicPath, 'index.html')))
     }
     if (isHoodiePath) {
-      return reply.continue()
+      return reply(fs.createReadStream(path.join(hoodiePublicPath, 'index.html')))
     }
 
     // Serve index.html
--- a/test/integration/handle-404-test.js
+++ b/test/integration/handle-404-test.js
@@ -13,7 +13,7 @@ var hapiPluginOptions = {
 
 require('npmlog').level = 'error'
 
-test('forward all requests that accept html to app', function (group) {
+test('forward requests to app', function (group) {
   group.test('send index.html on accept: text/html', function (t) {
     var server = new Hapi.Server()
     server.connection({port: 8090})
@@ -34,7 +34,7 @@ test('forward all requests that accept html to app', function (group) {
     })
   })
 
-  group.test('send a JSON 404 on anything but accept: text/html*', function (t) {
+  group.test('send a JSON 404 on accept: application/json', function (t) {
     var server = new Hapi.Server()
     server.connection({port: 8090})
     server.register(hapiPluginOptions, function (error) {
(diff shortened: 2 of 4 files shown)


jamietanna:
Note that this commit will _not_ work until the PR at https://github.com/hoodiehq/hoodie-store/pull/35 is merged.

jamietanna:
@gr2m, the latest commit added will return the custom 404 page (from https://github.com/hoodiehq/hoodie-store/pull/35) when we hit a CSS, JS, PNG or JPEG.
However, the original issue https://github.com/hoodiehq/hoodie/issues/595 isn't quite fixed; when we get a page not found, did we want it to go to the custom `404.html`, or to return the `hoodie/index.html` file?

gr2m:
I think I would rather tend to return the `hoodie/index.html` file, so we can use `history.pushState` in our `/hoodie` generic UI?

gr2m:
Hey @jamietanna, is this good to go from your side? Please let me know if you have any questions :)

jamietanna:
@gr2m once you've resolved the above question (above, regarding 200 vs 404) I'm happy for https://github.com/hoodiehq/hoodie-store/pull/35 then this to be merged.

gr2m:
@jamietanna okay looks all good, just one question: why do you put the `404` into `@hoodie/store` via  hoodiehq/hoodie-store#35, why note into `public/404.html` in this repository?

jamietanna:
That was to do with the fact that I saw `index.html` being served out of the store (https://github.com/hoodiehq/hoodie/pull/607/files#diff-43557d6060003ff6997c94943ed23a82R119) so I did it for this too. Not sure if that's the right approach?

gr2m:
above [the line you liked](https://github.com/hoodiehq/hoodie/pull/607/files#diff-43557d6060003ff6997c94943ed23a82R119) we check if the request path starts with `/hoodie/store`, which is why we serve the `index.html` from `@hoodie/store`. 

For this PR, let’s put the `404.html` in `hoodie`. We also recently decided to merge the static files from `@hoodie/store` and `@hoodie/account` into `hoodie` anyway, see https://github.com/hoodiehq/camp/issues/98 (you can work on this one next if you like ;) )

gr2m:
hey @jamietanna do you want to finish this up?

jamietanna:
Apologies @gr2m - see the latest commit, which adds in the `404.html` to `@hoodie`. _Before_ merge, I'd like to do a rebase just to fix the history, but only once you're happy with the changes.

server/plugins/public.js
```diff
@@ -95,13 +95,17 @@ function register (server, options, next) {
...
     var isAccountPath = /^\/hoodie\/account\//.test(request.path)
     var isAdminPath = /^\/hoodie\/admin\//.test(request.path)
     var isStorePath = /^\/hoodie\/store\//.test(request.path)
     var isHoodiePath = /^\/hoodie\//.test(request.path)
 
     // We only care about 404 for html requests...
     if (!is404 || !isHTML) {
+      if (is404 && isResource) {
+        return reply(fs.createReadStream(path.join(hoodiePublicPath, '404.html'))).code(404)
+      }
```
gr2m: I think we only need this now, or am I missing something?

```js
if (is404 && !isHTML) {
  return reply(fs.createReadStream(path.join(hoodiePublicPath, '404.html'))).code(404)
}
```

then you can remove `isResource` from line 98, too

test/integration/handle-404-test.js
```diff
@@ -53,20 +53,100 @@ test('forward all requests that accept html to app', function (group) {
     })
```
gr2m, line 54: line 49 should now be `t.match(response.payload, /<html/, 'response is HTML')` as we we return our custom 404.html page for non-html requests

server/plugins/public.js
```diff
@@ -114,6 +118,9 @@ function register (server, options, next) {
     if (isStorePath) {
       return reply(fs.createReadStream(path.join(storePublicPath, 'index.html')))
     }
+    if (is404 && isHoodiePath) {
+      return reply(fs.createReadStream(app))
+    }
```
gr2m: with the change above, we can do 

```js
if (isHoodiePath) {
  return reply(fs.createReadStream(path.join(hoodiePublicPath, 'index.html')))
}
```

and remove the `if (isHoodiePath) {` block below

gr2m (changes requested):
Okay I did a thorough review and had only few comments to simplify the code a bit and fix an error that has been there before (`reply(fs.createReadStream(app)` should really be `eply(fs.createReadStream(path.join(hoodiePublicPath, 'index.html')))` for `/hoodie/...` paths). 

Make sure to rebase on the master branch before cleaning up the commit history :) the coverage should stay at 100%, if not, you can run `npm test` and then `./node_modules/.bin/nyc report --reporter lcov` to generate a local coverage report into `coverage/lcov-report/index.html`. 

Great work, thanks 👍 Please let me know if you have any further questions

gr2m:
@jamietanna you still on this :) Anything I can help with?

jamietanna:
@gr2m sorry, yes - see the above question regarding the 404 page

gr2m:
> However, the original issue https://github.com/hoodiehq/hoodie/issues/595 isn't quite fixed; when we get a page not found, did we want it to go to the custom `404.html`, or to return the `hoodie/index.html` file?

if we get a page not found and the and the `Accept` header isn't `text/html` or there is a file extension that is not `.html`, return our 404.html

jamietanna:
Thanks @gr2m, I'll look into this tonight.

gr2m:
@jamietanna great, let me know if you have any remaining questions. Looking forward to ship this :)

jamietanna:
@gr2m there are a couple of comments I've made above. Please let me know what you think. The latest commits may have changed whether your points in https://github.com/hoodiehq/hoodie/pull/607/files/aef36eb9d85e56a9ad44e2040bf2be7e073eaad0#diff-43557d6060003ff6997c94943ed23a82 still hold.

server/plugins/public.js
```diff
@@ -95,7 +95,7 @@ function register (server, options, next) {
 
     var is404 = response.output.statusCode === 404
     var isHTML = /text\/html/.test(request.headers.accept)
-    var isResource = /(text\/css|application\/javascript|image\/png|image\/jpeg)/.test(request.headers.accept)
```
gr2m: wouldn’t this be simpler with `var isResource = !isHtml`?

gr2m:
@jamietanna sorry I don’t see your comments? I can have a final review today, I think this is good to merge, maybe some slight style-only changes :)

jamietanna:
@gr2m there are a couple of comments that can be seen via https://github.com/hoodiehq/hoodie/pull/607/files, and the others are displaying in this thread - they may not be showing as latest activity, mind

gr2m:
I only see my two comments at https://github.com/hoodiehq/hoodie/pull/607/files

1. https://github.com/hoodiehq/hoodie/pull/607/files#r91834776
2. https://github.com/hoodiehq/hoodie/pull/607/files#r91834950

Are these the ones you are refering to?

jamietanna:
@gr2m I believe it's maybe because they're via the Github Code Review functionality? See below what I can see

2017-01-07-115046_852x354_scrot
2017-01-07-115015_988x1565_scrot

gr2m:
@jamietanna ah yes that’s a bit of a confusing UI that came with GitHub’s review feature. See the yellow `Pending` next to your username? You did not finish your review. For that you have to click on the big green "Review Changes" dropdown, and select one of the options, probably "Comment" in your case

gr2m:
I would return the custom 404 page for all non-HTML files, including requests to JSON. What argument to see against that?

jamietanna:
Thanks @gr2m, that makes sense why you've not been able to see my comments - looks like I misread that!

I was thinking maybe if you're expecting JSON and get back HTML it may be confusing. Then again, I guess same with other resource files. I'll update the tests to reflect this and have the `404.html` shown for all non-HTML requests :)

test/integration/handle-404-test.js
jamietanna: Should be resolved as of latest commit.

server/plugins/public.js
jamietanna: I think this could be updated in the reverse with:

```diff
-    var isHTML = /text\/html/.test(request.headers.accept)
     var isResource = /(text\/css|application\/json|application\/javascript|image\/png|image\/jpeg)/.test(request.headers.accept) || /.(css|js|png|jpg)$/.test(request.path)
+    var isHTML = !isResource
```

jamietanna:
Looks like this is done functionality wise (although second opinion would be great!).

Things left to do:

- [ ] Squash history?
- [ ] Merge with master

gr2m (approved):
Great work 👍 I will clean up the commits and get it merged. Thanks for all your patience and thorough work 👏

jamietanna:
Thanks for the merge @gr2m! Out of interest, how come the integration tests weren't added to test that we're returning the `404.html` for resource files?

gr2m:
did something get lost you created? I did not remove anything, I only squashed your commits to follow our conventions, commiting all changes to the test files first, then the code for the implementation

jamietanna:
Apologies, I didn't realise there were two commits there - looks all good!

PR #626, opened 2016-10-31
Update CONTRIBUTING.md

dhruvaljain:


CONTRIBUTING.md
```diff
@@ -67,7 +67,7 @@ Example:
 
 ## Feature requests
 
-Feature requests are welcome. But take a moment to find out whether your idea
+Feature requests are always welcome. But take a moment to find out whether your idea
```
gr2m, line 70: This is just my opinion, wondering about others: I’m not sure we we need to stress that more than it is. More feature requests means more work for maintainers, it’s more important to me that people reflect on their idea and review the project than making it sound as if people would do us a favor if they send feature requests.

PR #627, opened 2016-10-31
Fix #609 - Hoodie default page not showing

### LINKED ISSUE #609: Hoodie default page not showing
I am running through the getting started guide and when I run the app without a public folder, I don't see the hoodie default page telling me that I need a `public` folder.

Also, when I navigate to any url e.g localhost:8080/home.html or any, it doesn't return a 404 or something.  Don't know if that's normal

minrwhite:
closes #609

### DIFF
--- a/cli/webroot-locator.js
+++ b/cli/webroot-locator.js
@@ -1,4 +1,5 @@
 module.exports = locateWebroot
+
 var fs = require('fs')
 var pathModule = require('path')
 
--- a/cli/webroot-locator.js
+++ b/cli/webroot-locator.js
@@ -1,0 +1,11 @@
+module.exports = locateWebroot
+var fs = require('fs')
+var pathModule = require('path')
+
+function locateWebroot (path) {
+  if (fs.existsSync(path) && fs.statSync(path).isDirectory()) {
+    return path
+  }
+
+  return pathModule.resolve(__dirname, '../public')
+}
--- a/test/unit/cli/options-test.js
+++ b/test/unit/cli/options-test.js
@@ -36,6 +36,8 @@ var packageJsonMock = {
   version: '1.0.0'
 }
 
+var mockWebrootLocator = simple.spy(function (arg) { return arg })
+
 function createCliOptionsProxy (yargsApi) {
   return proxyquire('../../../cli/options', {
     'npmlog': { warn: simple.spy() },
@@ -48,6 +50,7 @@ function createCliOptionsProxy (yargsApi) {
       }
     },
     './app-defaults': mockAppDefaults(),
+    './webroot-locator': mockWebrootLocator,
     'yargs': yargsApi,
     '../package.json': packageJsonMock
   })
@@ -96,6 +99,16 @@ test('config', function (group) {
 
     simple.restore(console, 'log')
     simple.restore(process, 'exit')
+    t.end()
+  })
+
+  group.test('public option', function (t) {
+    var fallbackValue = 'fallback-public'
+    mockWebrootLocator.returnWith(fallbackValue)
+
+    var options = getCliOptions(yargsApi)
+
+    t.equal(options.public, fallbackValue, 'uses the value returned from webroot locator')
     t.end()
   })
 
--- a/test/unit/cli/options-test.js
+++ b/test/unit/cli/options-test.js
@@ -36,7 +36,7 @@ var packageJsonMock = {
   version: '1.0.0'
 }
 
-var mockWebrootLocator = simple.stub(function (arg) { return arg })
+var mockWebrootLocator = simple.stub()
 
 function createCliOptionsProxy (yargsApi) {
   return proxyquire('../../../cli/options', {
--- a/test/unit/cli/options-test.js
+++ b/test/unit/cli/options-test.js
@@ -36,7 +36,7 @@ var packageJsonMock = {
   version: '1.0.0'
 }
 
-var mockWebrootLocator = simple.spy(function (arg) { return arg })
+var mockWebrootLocator = simple.stub(function (arg) { return arg })
 
 function createCliOptionsProxy (yargsApi) {
   return proxyquire('../../../cli/options', {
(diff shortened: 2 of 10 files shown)


cli/webroot-locator.js
```diff
@@ -0,0 +1,11 @@
+module.exports = locateWebroot
+var fs = require('fs')
```
gr2m, line 3: 🎨 can you please add an empty line below `module.exports = locateWebroot`

test/unit/cli/webroot-locator-test.js
```diff
@@ -0,0 +1,73 @@
...
+var statsMock = {
+  'isDirectory': simple.spy()
+}
+
+test('webroot locator', function (group) {
+  var webrootLocator = proxyquire('../../../cli/webroot-locator',
+    {fs: fsMock})
+
+  group.afterEach(function (done) {
+    fsMock.existsSync.actions.length = 0
```
gr2m: what is the `.actions.length = 0` for?

test/unit/cli/options-test.js
```diff
@@ -36,6 +36,8 @@ var packageJsonMock = {
   version: '1.0.0'
 }
 
+var mockWebrootLocator = simple.spy(function (arg) { return arg })
```
gr2m: I didn’t know that you can stub `simple.spy`s, too, but for coherence with other code, can you do `simple.stub()` here and in the other places please?

gr2m (changes requested):
looking great, thanks! Just a few comments

minrwhite:
actions.length = 0 is a way of resetting the stubbed behaviour, as spy.reset() only resets the logged calls. That way if we forget to change the behaviour between tests this should catch it and prevent a false positive.

Currently this fails, as public plugin currently determines the public webroot from the wrong config object path. Correcting this causes other tests to fail; I'll fix these tests as my next course of action.

gr2m:
but you don’t use `actions.length` anywhere else in the code? Am I missing something?

minrwhite:
It's shorthand for emptying an array. I can change this to be a bit more explicit if required?

minrwhite:
Rebased and pushed, ready for review, also pending yay/nay on the `actions.length = 0` array-emptying shorthand.

test/unit/cli/webroot-locator-test.js
minrwhite: Comment moved from below: It's shorthand for emptying an array. I can change this to be a bit more explicit if required?

test/unit/cli/options-test.js
```diff
@@ -36,6 +36,8 @@ var packageJsonMock = {
   version: '1.0.0'
 }
 
+var mockWebrootLocator = simple.stub(function (arg) { return arg })
```
gr2m: you can remove `function (arg) { return arg }`

test/unit/cli/webroot-locator-test.js
gr2m: okay I understand what `stub.actions` are now. I didn’t know one can chain `callbackWith`, `returnWith` and `throwWith`, that’s pretty cool! I would think that `stub.reset()` would remove these though, I’ll file an issue on simple-mock.

But please change the `actions.length = 0` to `actions = []`, it’s shorter and more readable

test/unit/cli/webroot-locator-test.js
```diff
@@ -0,0 +1,73 @@
...
+  'statSync': simple.stub()
+}
+
+var statsMock = {
+  'isDirectory': simple.stub()
+}
+
+test('webroot locator', function (group) {
+  var webrootLocator = proxyquire('../../../cli/webroot-locator',
+    {fs: fsMock})
```
gr2m: can you please change this to 

``` js
var webrootLocator = proxyquire('../../../cli/webroot-locator', {
  fs: fsMock
})
```

gr2m (changes requested):
okay final comments. I can also fix it up myself if you are busy?

minrwhite:
Actioned review changes, and rebased ready for review.

test/unit/cli/options-test.js
minrwhite: Indeed I can! Done.

test/unit/cli/webroot-locator-test.js
minrwhite: Changed. Agreed that that's clearer.

gr2m (approved):

gr2m:
beautiful work, thanks a ton

brunopedroso:
ooops.. should we check if the `server.paths` key is present? otherwise I'm forced to define it like

```
paths:{}
```

I'm using ember-hoodie...

gr2m:
it gets defaulted here: https://github.com/hoodiehq/hoodie/blob/047a8ff004538e68ccf2a89df33108fec505c3f3/server/index.js#L35-L40

brunopedroso:
yes, but only if I [dont define](https://github.com/hoodiehq/hoodie/blob/047a8ff004538e68ccf2a89df33108fec505c3f3/server/index.js#L26) `db.url`.

We're setting `db.url` here, and our app started to break on [this line](https://github.com/hoodiehq/hoodie/pull/627/files#diff-43557d6060003ff6997c94943ed23a82R11) just after I made the update. I had to define `paths:{}`...

Maybe that line could read like this?

```
var publicFolder = (options.config.paths && options.config.paths.public) || 'public'
```

minrwhite:
FWIW, I'd be tempted to resolve this by moving [this clause](/hoodiehq/hoodie/blob/047a8ff004538e68ccf2a89df33108fec505c3f3/server/index.js#L35) alongside [this one](/hoodiehq/hoodie/blob/047a8ff004538e68ccf2a89df33108fec505c3f3/server/index.js#L19), removing the 'data' key and letting [this](/hoodiehq/hoodie/blob/047a8ff004538e68ccf2a89df33108fec505c3f3/server/index.js#L41) handle the data key.

gr2m:
@brunopedroso ooops that is definitely wrong

The way it’s currently handled is not very elegant. I’ve a lot of work stuff to catch up with and won’t be able to look into this, but if one of you could send a PR that will fix @brunopedroso’s problem, I’m happy to review and merge

PR #635, opened 2016-11-09
fix(server): Resolve issue with unset paths object

minrwhite:
Resolve the issue brought to light after merging #627, whereby the
paths.public key is not set for the server if you are specifying a
DB url.


gr2m (approved):

gr2m:
Thanks 👍

PR #637, opened 2016-11-15
App-specific client/server plugins 

### LINKED ISSUE #98: Treat Hoodie apps like plugins: add support for client/, server/ and admin/ folders
Hoodie plugins are currently [work in progress](https://github.com/gr2m/milestones/issues/89), but they will be structured the same way the core modules are, see for example [hoodie-account](https://github.com/hoodiehq/hoodie-account) (Note that it does not have an `admin/` folder, which will be used in future for custom admin UIs).

The idea is that if you need custom server or client functionality, you are not forced to create a separate module for it. Instead, you can add your code directly to the `client/`, `server/` and `admin/` folders, just like you would do in plugins.

I thought about this for a while now and think it makes a lot of sense. It would help to avoid workarounds like [this one](https://github.com/hoodiehq/hoodie/pull/482#issuecomment-222335305)

minrwhite:
closes hoodiehq/discussion#98

* [x] If detected, the code at `hoodie/client` relative to the app root is to be bundled alongside hoodie's client code
* [x] If detected, the code at `hoodie/server` relative to the app root is to be loaded as a hapi plugin
* [x] Check modified time of plugins to build bundle as well as hoodie's internal client code
* [x] Update `README.md` to document hoodie's consumption of application-defined server/client code
* [x] (After merging) Open issue to document approach to better handling updated app client code [as described here](#discussion_r92874443) ▶️ https://github.com/hoodiehq/hoodie/issues/655

### DIFF
--- a/server/plugins/client/bundle.js
+++ b/server/plugins/client/bundle.js
@@ -1,33 +1,54 @@
 module.exports = bundleClient
 
 var fs = require('fs')
+var path = require('path')
 var parallel = require('async').parallel
+var requireResolve = require('../resolver')
+
+function checkModule (module) {
+  try {
+    requireResolve(module)
+    return true
+  } catch (err) {
+    if (err.code !== 'MODULE_NOT_FOUND') {
+      throw err
+    }
+    return false
+  }
+}
 
 /**
- * we compare the mtime (modified time) for the Hoodie client and the target
- * bundle file. If the latter does not exist or Hoodie client was modified
- * more recently, we read out the client and add the `new Hoodie()` init code,
- * otherwise we simply read out the bundle file.
+ * we compare the mtime (modified time) for the last modified dependency of the
+ * Hoodie client and the target bundle file. If the latter does not exist or
+ * any dependency was modified more recently, we read out the client and add
+ * the `new Hoodie()` init code, otherwise we simply read out the bundle file.
  *
  * This optimisation is in preparation for plugins. Plugins can extend the
  * client, so we need to browserify on-the-fly to avoid dependency duplication,
  * and avoiding unneeded bundling with browserify saves a significant time.
  */
 function bundleClient (hoodieClientPath, bundleTargetPath, config, callback) {
-  parallel([
+  var plugins = [
+    path.resolve('hoodie/client')
+  ].filter(checkModule)
+  var getPluginsModifiedTimes = plugins.map(function (pluginPath) {
+    return getModifiedTime.bind(null, requireResolve(pluginPath))
+  })
+
+  parallel(getPluginsModifiedTimes.concat([
     getModifiedTime.bind(null, hoodieClientPath),
     getModifiedTime.bind(null, bundleTargetPath)
-  ], function (error, results) {
+  ]), function (error, results) {
     /* istanbul ignore if */
     if (error) {
       return callback(error)
     }
 
-    var sourceTime = results[0]
-    var targetTime = results[1]
+    var targetTime = results.pop()
+    var sourceTime = Math.max.apply(null, results)
     var hasUpdate = sourceTime > targetTime
 
-    var get = hasUpdate ? buildBundle.bind(null, config) : fs.readFile.bind(null, bundleTargetPath)
+    var get = hasUpdate ? buildBundle.bind(null, config, plugins) : fs.readFile.bind(null, bundleTargetPath)
 
     get(function (error, buffer) {
       if (error) {
@@ -50,7 +71,7 @@ function getModifiedTime (path, callback) {
   })
 }
 
-function buildBundle (config, callback) {
+function buildBundle (config, plugins, callback) {
   var ReadableStream = require('stream').Readable
   var browserify = require('browserify')
   var stream = new ReadableStream()
@@ -70,6 +91,9 @@ function buildBundle (config, callback) {
   hoodieBundleSource += '  PouchDB: require("pouchdb-browser")\n'
   hoodieBundleSource += '}\n'
   hoodieBundleSource += 'module.exports = new Hoodie(options)\n'
+  plugins.forEach(function(pluginPath) {
+    hoodieBundleSource += '  .plugin(require("' + pluginPath + '"))\n'
+  })
 
   stream.push(hoodieBundleSource)
   stream.push(null)
--- a/server/plugins/resolver.js
+++ b/server/plugins/resolver.js
@@ -1,0 +1,1 @@
+module.exports = require.resolve
--- a/README.md
+++ b/README.md
@@ -149,6 +149,43 @@ option                    | default      | description
 **account**               | `{}`         | [Hoodie Account Server](https://github.com/hoodiehq/hoodie-account-server/tree/master/plugin#options) options. `account.admins`, `account.secret` and `account.usersDb` are set based on `db` option above.
 **store**                 | `{}`         | [Hoodie Store Server](https://github.com/hoodiehq/hoodie-store-server#options) options. `store.couchdb`, `store.PouchDB` are set based on `db` option above. `store.hooks.onPreAuth` is set to bind user authentication for Hoodie Account to Hoodie Store.
 
+## Extending Hoodie for your app
+
+You can add custom behavior to your hoodie-based application in a few ways:
+
+1. The server-side functionality can be extended by adding a module constructed as a [hapi plugin](http://hapijs.com/tutorials/plugins)
+ to the path `hoodie/server`. E.g:
+ 
+```js
+exports.register = function (server, options, next) {
+  // server-relevant code here
+  next();
+}
+
+exports.register.attributes = {
+  'name': 'my-app'
+}
+```
+ 
+2. The client-side functionality can be extended by adding a module that can be interpreted as a [hoodie plugin](https://github.com/hoodiehq/hoodie-client#hoodieplugin)
+ to the path `hoodie/client`, which by default will be built into the client.js package by browserify. 
+ 
+```js
+module.exports = {
+  demonstratePlugin: function () { // will be attached as hoodie.demonstratePlugin()
+    // client-relevant code here
+  }
+}
+```
+
+Both of the above paths will be imported as if Node modules, so the path to the entry file for each module can be
+`hoodie/{server,client}.js` or `hoodie/{server,client}/index.js`.
+   
+Note that for now only the file modification time of the entry file of the client module is used to determine the
+freshness of the client bundle and therefore whether it should be rebuilt on the first request after a server restart.
+If your client code stretches to many files, you may need to update the `mtime` of the client entry file by using
+`touch` or a similar method before restarting your server to ensure that the client bundle is rebuilt.
+   
 ## Testing
 
 Local setup
--- a/cli/parse-options.js
+++ b/cli/parse-options.js
@@ -25,7 +25,8 @@ function parseOptions (options) {
       data: options.data,
       public: options.public
     },
-    db: {}
+    db: {},
+    inMemory: Boolean(options.inMemory)
   }
 
   log.level = config.loglevel
@@ -36,9 +37,6 @@ function parseOptions (options) {
 
   if (options.dbUrl) {
     config.db.url = options.dbUrl
-  }
-  if (options.inMemory) {
-    config.inMemory = true
   }
 
   return config
--- a/server/index.js
+++ b/server/index.js
@@ -16,14 +16,12 @@ var registerPlugins = require('./plugins')
 
 function register (server, options, next) {
   options = _.cloneDeep(options)
-  if (!options.db) {
-    options.db = {}
-  }
-  if (!options.paths) {
-    options.paths = {
+  _.defaultsDeep(options, {
+    db: {},
+    paths: {
       public: 'public'
     }
-  }
+  })
 
   // mapreduce is required for `db.query()`
   PouchDB.plugin(require('pouchdb-mapreduce'))
--- a/server/plugins/index.js
+++ b/server/plugins/index.js
@@ -1,6 +1,20 @@
 module.exports = registerPlugins
 
 var log = require('npmlog')
+var path = require('path')
+var requireResolve = require('./resolver')
+
+function checkModule (module) {
+  try {
+    requireResolve(module)
+    return true
+  } catch (err) {
+    if (err.code !== 'MODULE_NOT_FOUND') {
+      throw err
+    }
+    return false
+  }
+}
 
 function registerPlugins (server, config, callback) {
   var options = {
@@ -9,17 +23,25 @@ function registerPlugins (server, config, callback) {
   var hapiPlugins = [
     require('inert')
   ]
+
   var localPlugins = [
-    require('./client'),
-    require('./logger'),
-    require('./maybe-force-gzip'),
-    require('./public')
-  ].map(function (register) {
-    return {
-      options: options,
-      register: register
-    }
-  })
+    './client',
+    './logger',
+    './maybe-force-gzip',
+    './public'
+  ]
+    .concat(
+  [
+    path.resolve('hoodie/server')
+  ]
+    .filter(checkModule)
+    )
+    .map(function (register) {
+      return {
+        options: options,
+        register: require(register)
+      }
+    })
 
   log.silly('hapi', 'Registering internal plugins')
   server.register(hapiPlugins.concat(localPlugins), function (error) {
--- a/server/plugins/public.js
+++ b/server/plugins/public.js
@@ -8,7 +8,7 @@ var fs = require('fs')
 var path = require('path')
 
 function register (server, options, next) {
-  var publicFolder = options.config.paths.public || 'public'
+  var publicFolder = options.config.paths.public
   var app = path.join(publicFolder, 'index.html')
   var hoodieVersion
   try {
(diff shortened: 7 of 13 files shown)


gr2m:
I’ve seen that you’ve already put it into this pull request, but I’m now definitely convinced that we should use `/hoodie` as the main folder and then put `client/`, `server/` and other things in future inside there, to avoid conflicts in future.

And when we do that, I wonder if we should allow to configure the client & server paths at all? I’d argue no, not only is it good to enforce this convention, but also will it help us when we implement 3rd party plugins, as we will always know from where to load the module’s client & server code, without finding a configuration first.

what do you think? How are things going in general?

minrwhite:
I'm definitely happy to enforce the /hoodie/{public/,client/,server/} architecture for hoodie plugins proper, but believe that the equivalent paths within the top-level app should still be configurable, as to me they have a use case as app-level configured values (within the app's `package.json`) but just not as CLI options (why would you need to change this on a run-by-run basis) - does this make sense to you?

In general I haven't had a chance to touch this since last Tuesday; I'm giving it another look over today.

gr2m:
We us [rc](https://www.npmjs.com/package/rc) for options so they can come from a `.hoodierc` file, environment variables, CLI arguments or the app-specific defaults from the package.json.

I would like to enforce this for top-level apps as well, as apps can be used as plugins themselves, and if the logic will be put in other directories, that will break. That’s not decided yet though, it’s just where I’m standing right now, and we had architectures in the past where this would apply (we run plugins in separate node processes or altogether in one dependent on the environment). 

Can you elaborate on "as to me they have a use case as app-level configured values"? What’s your use case to have these paths being configurable?

minrwhite:
I must have been thinking about if anyone wanted to use the `hoodie/` path for their own purposes, it would be good to avoid an overlap by some mechanism. This is probably extremely edge-case though and better worked around.

I understand the value of having apps also be usable as plugins/modules in their own right (coming from a PHP knowledge base, this is prevalent in Zend Framework, for example) in which case enforcing the relative path from each module's root to be standard makes complete sense. I'm happy to proceed on this.

gr2m:
okay, let’s enforce the paths for now then and remove the configuration

gr2m:
@benwhite-deltas how are things going? I had quite a few updates in the past two weeks, and among other things, I think the way the bundling of the Hoodie client works will make it pretty simple to load client plugins, see https://github.com/hoodiehq/hoodie/blob/deac802a5b4ab31d7ba99d5b25150b9a696fbfbd/server/plugins/client/bundle.js#L61-L74

Instead of 

```js
module.exports = new Hoodie(options)
```

we can do

```js
var hoodie = new Hoodie(options)
hoodie.plugin(require('./hoodie/client')
module.exports = hoodie
```

if `hoodie/client/index.js` is present. Makes sense?

For the server plugins, we need to load the custom plugins after we register `@hoodie/server`:
https://github.com/hoodiehq/hoodie/blob/deac802a5b4ab31d7ba99d5b25150b9a696fbfbd/server/index.js#L78

Let me know if you have any questions, it’d be great to move forward with this, it’s a huge milestone for us :)

gr2m:
make sure to rebase on latest `master`, there should be no conflicts. Maybe do `rm -rf node_modules && npm install` because there have been lots of changes

minrwhite:
Hi @gr2m, I was working on a contract over the past few weeks, but looking to be on this again from tomorrow.

Agreed, that plugin loading method looks pretty schweet!

gr2m:
Hey @benwhite-deltas, any progress? No pressure, just checking in. Please let me know if there is anything I can help with :)

minrwhite:
Hi @gregor, just picked this up again. Looking at it, although it rebases cleanly, the vast majority of the PR will no longer be required as it's pretty much option handling.

If we're resolving the `client` and `server` dirs relative to the app root, my plan is to pass the app root as determined from `process.cwd()` in `cli/index.js` into the server registration function as an option. This won't be linked to a `yargs`/`rc` option as we want to enforce the paths, but will be allowed when calling server as a registration function to allow the app root to be set when calling hoodie as a hapi plugin.

From that point within `server` it's pretty much determining if the `server` and/or `client` dirs resolve relative to the app root, and inserting them as appropriate within either the hapi plugin stack or browserify queue respectively.

Does that all make sense?

gr2m:
> Hi @gregor, just picked this up again. Looking at it, although it rebases cleanly, the vast majority of the PR will no longer be required as it's pretty much option handling.

I’m sorry that it’s all so much extra work :( 

> If we're resolving the client and server dirs relative to the app root, my plan is to pass the app root as determined from process.cwd() in cli/index.js into the server registration function as an option. This won't be linked to a yargs/rc option as we want to enforce the paths, but will be allowed when calling server as a registration function to allow the app root to be set when calling hoodie as a hapi plugin.

yeah that should work 👍 

> From that point within server it's pretty much determining if the server and/or client dirs resolve relative to the app root, and inserting them as appropriate within either the hapi plugin stack or browserify queue respectively.

> Does that all make sense?

Yes it all makes sense

minrwhite:
Not a problem, good code should be written to be deleted later :smiley:

In the current state of this PR, the server code should be loading in - now to add the client code.

minrwhite:
@gregor Extra to the above API, because hoodie's client `plugin` method has a fluent interface, we don't even need to assign it to a variable - we can simply add plugins on to the end, e.g:
```js
    module.exports = new Hoodie(options)
      .plugin(require(appRoot + '/hoodie/client'))
      .plugin(require(anotherPlugin))
```

cli/hoodie-defaults.js
```diff
@@ -9,6 +9,7 @@ function getHoodieDefaults () {
     dbUrl: undefined,
     inMemory: false,
     loglevel: 'warn',
-    url: undefined
+    url: undefined,
+    applicationRoot: undefined
```
gr2m: I’m wondering if we really need the `applicationRoot` option right now? I guess we would set it if I want to start a hoodie server in another folder, but right now I don’t see a use case for it? Do you?

gr2m: Oh and just so I don’t forget, in case we agree that we should add it, let’s also document it in the README :)

server/plugins/client/bundle.js
```diff
@@ -14,20 +28,27 @@ var parallel = require('async').parallel
...
+  ]), function (error, results) {
     /* istanbul ignore if */
     if (error) {
       return callback(error)
     }
 
-    var sourceTime = results[0]
-    var targetTime = results[1]
+    var sourceTime = Math.max.apply(null, results.slice(0, -1))
+    var targetTime = results[results.length - 1]
```
gr2m: just so I understand this correctly, before we only had one source and one target. Now we can have multiple sources, so we need an array which we create of the results excluding the last item in  the array?

I’m not sure if this is better, but this is how my brain would have implemented this, what do you think?

```js
var targetTime = results.pop()
var sourceTime = Math.max.apply(Math, results)
```

server/plugins/client/bundle.js
```diff
@@ -14,20 +28,27 @@ var parallel = require('async').parallel
  * and avoiding unneeded bundling with browserify saves a significant time.
  */
 function bundleClient (hoodieClientPath, bundleTargetPath, config, callback) {
-  parallel([
+  var plugins = [
+    path.resolve(config.paths && config.paths.applicationRoot || '', 'hoodie/client')
+  ].filter(checkModule)
+  var getPluginsModifiedTimes = plugins.map(function (pluginPath) {
+    return getModifiedTime.bind(null, requireResolve(pluginPath))
+  })
```
gr2m, line 36: `getPluginsModifiedTimes` sounds like a method, but it’s an array with values, right? Can we rename it to `pluginsModifiedTimes`?

If I understand it correctly, we get the entry JavaScript file for the app-specific client plugin, usually that would be `hoodie/client/index.js` or `hoodie/client.js`. 

The problem that we have is that while the client bundle will always be one file, the app-specific plugin (and 3rd party plugins in future) will most be spread around multiple files. I don’t have an idea how take this into account right now, I think this might end up being a very hard problem to solve :/ Maybe it’s something that browserify itself can do, with a custom plugin or something?

Hmmmmm 🤔

gr2m (commented):
this looks really, really good, I think we are very close! Just left a few comments :)

cli/hoodie-defaults.js
minrwhite: The use cases I'd thought of and was trying to cater for, where the output from `process.cwd()` could not necessarily be relied on:

* if an application module that defined hoodie's behaviour was being included as a dependency of another module
* if an application defined it's own cli application available via the `PATH`

Note that the option is not exposed on the CLI, it's only included in `getHoodieDefaults()` to prevent it being filtered out before the options are passed to the server.

Within `server/index.js` it's defaulted to `process.cwd()` anyway, so if these use cases can be discounted, I don't have a problem with removing it for now.

Let me know either way, I'll remove it if we don't need it, or update the README if we do.

server/plugins/client/bundle.js
minrwhite: > `getPluginsModifiedTimes` sounds like a method, but it’s an array with values, right? Can we rename it to `pluginsModifiedTimes`?

No, it's an array (of one at the moment, but with the intention that it is ready to expand to accept custom plugins) of bound instances of the `getModifiedTime` function, which are then run in parallel with the other bound instances. I did debate over the naming of that one - how do you name an array of functions? Maybe taking a cue from FTP, how about `mgetPluginsModifiedTimes`?

> If I understand it correctly, we get the entry JavaScript file for the app-specific client plugin, usually that would be `hoodie/client/index.js` or `hoodie/client.js`. 

> The problem that we have is that while the client bundle will always be one file, the app-specific plugin (and 3rd party plugins in future) will most be spread around multiple files. I don’t have an idea how take this into account right now, I think this might end up being a very hard problem to solve :/ Maybe it’s something that browserify itself can do, with a custom plugin or something?

I think hooking into the browserify API may indeed be the way to go - browserify uses [module-deps](https://github.com/substack/module-deps) internally to calculate the dependency tree, and allows emitting the output from this as an event on it's internal [pipeline](https://github.com/substack/node-browserify#bpipeline). If we save the output from this as a JSON file alongside the bundle, we can load this in, recheck all of the dependencies for modifications, and regenerate the dependency tree each time the server is started, with little extra cost due to it being work browserify is doing anyway. Sound like a plan?

server/plugins/client/bundle.js
minrwhite: > just so I understand this correctly, before we only had one source and one target. Now we can have multiple sources, so we need an array which we create of the results excluding the last item in the array?

Yes, that pretty much sums it up.

> ```js
> var targetTime = results.pop()
> var sourceTime = Math.max.apply(Math, results)
> ```

D'oh! Yep this is more efficient. I think I was trying to getaway without mutating the argument, but it is throwaway data after all. I'll make this change.

minrwhite: I'll also update the comment at the head of the function as part of this.

cli/hoodie-defaults.js
gr2m: > if an application module that defined hoodie's behaviour was being included as a dependency of another module

If we enforce the folder structure so that client/server plugins can be required with `require(module-name-here/hoodie/client)` and `require(module-name-here/hoodie/server)` then we should be good. If the an app is included as a dependency, that the plugins should also be loaded as third party plugins, right?

> if an application defined it's own cli application available via the PATH

I can’t follow you here. What do you mean with "if an application defined its own CLI application"?

When in doubt, I would rather keep options to a minimum and only add them if we there are specific use cases. Because removing options is much harder than adding them later on

server/plugins/client/bundle.js
gr2m: > No, it's an array [...] of bound instances of the getModifiedTime function

Ooops you are totally right, no idea how I could miss that. The name is good then, sorry for the confusion :)

> I think hooking into the browserify API may indeed be the way to go

Yeah I was thinking about saving the output of a JSON tree somehow. But it sounds fairly complicated, I wouldn’t mind to leave things simple as they are right now and document its limitation, with a link to an issue that describes it in detail and what the approach would be to fix it, so that we can work on this separately.

I think your plan sounds great :) A few more thoughts for the bundled hoodie.js client: 

1. Sometimes people  don’t need the bundling at all, because they bundle the client manually into their frontend application assets, so it would be nice to disable it altogether and return a helpful error message
2. The bundling should only be done in development. For deployment, we still need to figure out a way that would let us bundle the client and then make it part of the deployed server. I don’t yet know how to do it, so just want to put this thought out there :)

I think it makes sense to get your work so far merged first as it unblocks other important work, and then look into a proper implementation of the client bundle cache?

cli/hoodie-defaults.js
minrwhite: > If the an app is included as a dependency, that the plugins should also be loaded as third party plugins, right?

I think It depends how we're introducing the app-come-module to hoodie, whether the consuming app will be using whatever API we eventually decide on for registering plugins with hoodie (which is what I think you're talking about here), or the consuming app doesn't know about hoodie at all and the app-come-module is loading hoodie itself. But we're probably not at the level of decoupling from Hapi to make the latter option a simple prospect.

> I can’t follow you hear. What do you mean with "if an application defined its own CLI application"?

In the same way that the now-deprecated hoodie-cli used to, if a consuming app has an executable globally installed via npm, it could potentially be run from any directory, making `process.cwd()` unreliable. We could forbid this usage for now though and consider reinstating this option if someone genuinely butts up against this use-case.

server/plugins/client/bundle.js
minrwhite: > I wouldn’t mind to leave things simple as they are right now and document its limitation, with a link to an issue that describes it in detail and what the approach would be to fix it, so that we can work on this separately.

As you correctly interpret above, this code will get the path to the entry file of the client code, so we can place in the `README.md` section regarding how to write code for an app that is consumed by browserify and incorporated into the bundle ([nts](# "note to self"): which I should write as a final step for this PR) that you need to `touch` or otherwise alter the mtime on the entry file for the client code for it to be reconsumed, with a link to the issue-to-be-created you mention.

I don't think this is too much for someone at the moment, especially as the bundle is only re-built on a server restart for now due to the clever cache-by-Promise mechanism in `server/plugins/client/bundle-handler-factory.js`

> A few more thoughts for the bundled hoodie.js client: [...]

I think it's worth raising these as separate issues in addition to the above one, to at least help categorise thoughts on the separate topics.

minrwhite: Actually, I'm not sure where I'd put this documentation. There doesn't seem to be a natural fit for it in `README.md`. I think adding a tutorial at [http://docs.hood.ie/camp/tutorials](http://docs.hood.ie/camp/tutorials) would be prudent, but otherwise there doesn't seem to be a formal description of how you should arrange a hoodie application. Should there be a new section in `README.md` for application layout detailing `public` (or otherwise depending on the cli/plugin option), `hoodie/server` and `hoodie/client`, or should this information be something/somewhere else entirely?

cli/hoodie-defaults.js
gr2m: yes to both, we can’t decouble Hoodie from Hapi right now, and we don’t need to worry about a global `hoodie-cli` installation at this point.

server/plugins/client/bundle.js
gr2m: How about we add a section above `## Testing` and call it `## Extending Hoodie for your app` or something similar? 

It will also go into our docs, but that is another construction site :D

minrwhite:
Rebased onto latest master

server/plugins/resolver.js
```diff
@@ -0,0 +1 @@
+module.exports = require.resolve
```
gr2m, line 1: I wonder if it’s worth it to put this into it’s own file? I think we use `require.resolve` directly in other modules.

Also, all files in `server/plugins/` are valid hapi plugins, while this exposes an API. It’s just a little nitpick, I’d merge it in as is, we can also refactor it later :) But please let me know what your thinking is behind this file

gr2m:
Outstanding work 👏 

There is one line not fully covered by tests now:
[](https://github.com/benwhite-deltas/hoodie/blob/d5c2e154cfe7558b3bdcc556c995080a2cea8ab0/server/plugins/public.js#L11)

You can see it when you run `npm test` at the end of the output. If you want the HTML report that you can browse locally with your browser, run `./node_modules/.bin/nyc report --reporter=lcov` and open `coverage/lcov-report/index.html`.

If you’re to busy or off for well deserved vacation now, I’d be happy to fix this for you, just let me know

gr2m:
Hey Benjamin, I’ll finish this up as other PRs are piling up 👍 this is so awesome, such a huge milestone, thanks so much for all your work

gr2m (approved):
funk yeah 🙌

minrwhite:
@gr2m Thanks for progressing this, I was on a family holiday. I'm very glad to see this merged!

server/plugins/resolver.js
minrwhite: My thinking is that for this particular use of `require.resolve`, we expect to sometimes get an error (with code `MODULE_NOT_FOUND`), rather than a correctly resolved module. In order to unit test correctly for this case, we need to stub require.resolve to return an error predictably, hence why I've separated this out to be included as a module rather than use the inbuilt `require.resolve` directly.

PR #638, opened 2016-11-16
In Progress: changed all options.db to options.PouchDB in lines 19-59

jeanhl:
Next step is to adapt server-test.js

https://github.com/hoodiehq/camp/issues/51

### DIFF
--- a/server/index.js
+++ b/server/index.js
@@ -16,8 +16,8 @@ var registerPlugins = require('./plugins')
 
 function register (server, options, next) {
   options = _.cloneDeep(options)
-  if (!options.db) {
-    options.db = {}
+  if (!options.PouchDB) {
+    options.PouchDB = {}
   }
   if (!options.paths) {
     options.paths = {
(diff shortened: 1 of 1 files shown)


server/index.js
```diff
@@ -28,33 +28,33 @@ function register (server, options, next) {
   // mapreduce is required for `db.query()`
   PouchDB.plugin(require('pouchdb-mapreduce'))
 
-  if (!options.db.url) {
+  if (!options.PouchDB.url) {
```
gr2m, line 31: it’s more complicated than that. `options.db` is an object that you would pass into the `PouchDB` constructor, like this:

``` js
var db = new PouchDB(options.db)
```

But instead of passing in options, we want to pass in the constructor itself, does that make sense?

gr2m (changes requested):

gr2m:
@jeanhl do you want to continue working on this? I’d be happy to help if you have any questions?

jeanhl:
I'm planning on revisiting this issue but I don't think I can do it in a
timely manner. I'll need to read the code and understand what's the need
first.

I thought it was a simple replacement from options.db to options.PouchDB
but it seems more complicated than that. Polina thinks I need to write new
tests for the code.

gr2m:
> Polina thinks I need to write new tests for the code.

yes, most likely. Don’t worry if you can’t make it, just let us know, but I’m happy to help you get trough it, too :)

gr2m:
hey @jeanhl I’ll close this pull request in favor of #651, but please let me know if you’d like to work on something else, ok? I’d be happy to help find something and help you get started :)

PR #642, opened 2016-11-29
feat: Modified setup.js not to overide "start" script in package.json

Taekyoon:
Added a condition to check if packageJson.scripts["start"] has a value and blocked to override this value. Instead, overriding the "start" value, Added a statement to store this value to "start-hoodie" script in package.json

### DIFF
--- a/bin/setup.js
+++ b/bin/setup.js
@@ -20,7 +20,7 @@ if (saveRequested) {
     log.info('setup', 'start script already set to "' + packageJson.scripts[ 'start' ] +
     ', you can start hoodie with "npm run start-hoodie" instead')
 
-    packageJson.scripts[ 'start-hoodie' ] = packageJson.scripts[ 'start' ]
+    packageJson.scripts[ 'start-hoodie' ] = 'hoodie'
   } else {
     packageJson.scripts[ 'start' ] = 'hoodie'
   }
--- a/bin/setup.js
+++ b/bin/setup.js
@@ -15,7 +15,15 @@ if (saveRequested) {
   var packageJson = require(path.join(pathToAppRoot, 'package.json'))
 
   packageJson.scripts = packageJson.scripts || {}
-  packageJson.scripts[ 'start' ] = 'hoodie'
+
+  if (packageJson.scripts[ 'start' ]) {
+    log.info('setup', 'start script already set to "' + packageJson.scripts[ 'start' ] +
+    ', you can start hoodie with "npm run start-hoodie" instead')
+
+    packageJson.scripts[ 'start-hoodie' ] = packageJson.script[ 'start' ]
+  } else {
+    packageJson.script[ 'start' ] = 'hoodie'
+  }
 
   var newPackageJson = JSON.stringify(packageJson, null, 2)
 
--- a/bin/setup.js
+++ b/bin/setup.js
@@ -20,9 +20,9 @@ if (saveRequested) {
     log.info('setup', 'start script already set to "' + packageJson.scripts[ 'start' ] +
     ', you can start hoodie with "npm run start-hoodie" instead')
 
-    packageJson.scripts[ 'start-hoodie' ] = packageJson.script[ 'start' ]
+    packageJson.scripts[ 'start-hoodie' ] = packageJson.scripts[ 'start' ]
   } else {
-    packageJson.script[ 'start' ] = 'hoodie'
+    packageJson.scripts[ 'start' ] = 'hoodie'
   }
 
   var newPackageJson = JSON.stringify(packageJson, null, 2)


varjmes:
This is for issue #592

bin/setup.js
```diff
@@ -15,7 +15,15 @@ if (saveRequested) {
   var packageJson = require(path.join(pathToAppRoot, 'package.json'))
 
   packageJson.scripts = packageJson.scripts || {}
-  packageJson.scripts[ 'start' ] = 'hoodie'
+
+  if (packageJson.scripts[ 'start' ]) {
+    log.info('setup', 'start script already set to "' + packageJson.scripts[ 'start' ] +
+    ', you can start hoodie with "npm run start-hoodie" instead')
+
+    packageJson.scripts[ 'start-hoodie' ] = packageJson.script[ 'start' ]
```
varjmes: Correct me if I am wrong, but:

1. Sometimes you say `packageJson.scripts`, other times you say `packageJson.script`. I think it should always be `packageJson.scripts`?
2. Why are we setting `start-hoodie` to `start`. That feels like the opposite of what we want to do. Shouldn't we be setting `start-hoodie` to `hoodie`?
3. I think there might be an opportunity to write a test for this, somewhere in our unit tests, to prove that this works?

Wondering if @gr2m has any thoughts on my comments above.
Thanks so much for starting this PR ⭐️

varjmes (changes requested):

bin/setup.js
gr2m: Correct me if I am wrong, but:

> 1. Sometimes you say `packageJson.scripts`, other times you say `packageJson.script`. I think it should always be `packageJson.scripts`?

👍👁:clap:  great catch! `packageJson.script` should be `packageJson.scripts`

> 2. Why are we setting `start-hoodie` to `start`. That feels like the opposite of what we want to do. Shouldn't we be setting `start-hoodie` to `hoodie`?

Correct. I would suggest to replace line 23 with `packageJson.scripts[ 'start-hoodie' ] = 'hoodie'`

> 3. I think there might be an opportunity to write a test for this, somewhere in our unit tests, to prove that this works?

That’s my fault, I asked @Taekyoon to look into although this part had no tests yet. I’d suggest we test it locally to make sure it works, and add tests in a follow up pull request? But @Taekyoon if you like and feel like it, you can add tests now already, I’m happy to help!

bin/setup.js
```diff
@@ -15,7 +15,15 @@ if (saveRequested) {
...
+  } else {
+    packageJson.script[ 'start' ] = 'hoodie'
```
gr2m: as Charlotte mentioned above, this should be `packageJson.scripts[ 'start' ] = 'hoodie'`

gr2m (changes requested):

bin/setup.js
varjmes: That sounds good to me!
Glad I was on the right page here, too :)

Taekyoon: Oh! I think something mistook on the code. I should have written `packageJson.scripts`, instead `packageJson.script`. I'll change this code. And @gr2m, can you let me know how to build `setup.js` test? I've tested in other projects but they already set to run. I tried to find related test codes, but it doesn't exist.

bin/setup.js
```diff
@@ -15,7 +15,15 @@ if (saveRequested) {
   var packageJson = require(path.join(pathToAppRoot, 'package.json'))
 
   packageJson.scripts = packageJson.scripts || {}
-  packageJson.scripts[ 'start' ] = 'hoodie'
+
+  if (packageJson.scripts[ 'start' ]) {
+    log.info('setup', 'start script already set to "' + packageJson.scripts[ 'start' ] +
+    ', you can start hoodie with "npm run start-hoodie" instead')
+
+    packageJson.scripts[ 'start-hoodie' ] = packageJson.scripts[ 'start' ]
```
Taekyoon: @gr2m @Charlotteis Now I fixed wrong named variables. And do I need to change line 23 to `packageJson.scripts[ 'start-hoodie' ] = 'hoodie'`? I'm not sure I should change it.

bin/setup.js
Taekyoon: @gr2m @Charlotteis Now I fixed wrong named variables. And do I need to change line 23 to `packageJson.scripts[ 'start-hoodie' ] = 'hoodie'`? I'm not sure I should change it.

bin/setup.js
gr2m: this should be `packageJson.scripts[ 'start-hoodie' ] = 'hoodie'`, so that we can start the server with `npm run start-hoodie`. The current code sets it to whatever `npm start` does right now, which is not what we want :)

Taekyoon: Got it!

gr2m (approved):

gr2m:
looks good to me now, thanks for bearing with us @Taekyoon 👍 
Okay for you @Charlotteis to merge now? I’ll help Taekyoon to add tests in a follow up pull request

varjmes (approved):

varjmes:
sorry this took so long, just started a new job!

gg team!

PR #648, opened 2016-12-04
feat: Default url option to location.origin in bundle

sdvg:
Closes hoodiehq/hoodie-client#105. Closes https://github.com/hoodiehq/hoodie/pull/649

This belongs to hoodiehq/hoodie-client#109. The issue required changes in both repositories.

### DIFF
--- a/package.json
+++ b/package.json
@@ -13,7 +13,7 @@
   "dependencies": {
     "@hoodie/account": "^2.0.0",
     "@hoodie/admin": "^1.0.1",
-    "@hoodie/client": "^6.0.0",
+    "@hoodie/client": "^7.0.0",
     "@hoodie/server": "^21.1.0",
     "@hoodie/store": "^1.0.1",
     "async": "^2.0.0",
--- a/server/plugins/client/bundle.js
+++ b/server/plugins/client/bundle.js
@@ -65,9 +65,19 @@ function buildBundle (config, hoodieClientPath, callback) {
       return callback(error)
     }
 
-    var options = config.url ? '{url: "' + config.url + '"}' : ''
-    var initBuffer = Buffer('\n\nhoodie = new Hoodie(' + options + ')')
+    var initScript = '\n\n'
+    if (config.url) {
+      initScript += 'var hoodieOptions = {url: "' + config.url + '"}\n'
+    }
+    else {
+      initScript += 'var hoodieOptions = {}\n'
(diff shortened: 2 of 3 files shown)


server/plugins/client/bundle.js
```diff
@@ -65,9 +65,19 @@ function buildBundle (config, hoodieClientPath, callback) {
       return callback(error)
     }
 
-    var options = config.url ? '{url: "' + config.url + '"}' : ''
-    var initBuffer = Buffer('\n\nhoodie = new Hoodie(' + options + ')')
+    var initScript = '\n\n'
```
gr2m, line 68: this won’t hurt, but am curious: why not just `var initScript = ''`?

gr2m:
@sdvg [Hoodie Client v6 is released](https://github.com/hoodiehq/hoodie-client/releases/tag/v6.0.0), can you please update the `package.json` file as part of this PR?

server/plugins/client/bundle.js
sdvg: Only to have some space between the client and initialization code. It also was the same way before :)

sdvg:
@gr2m Isn't it already in master? https://github.com/hoodiehq/hoodie/blob/master/package.json#L16

gr2m:
ooops yes, sorry, we have to wait for v7, currently waiting for travis https://travis-ci.org/hoodiehq/hoodie-client/builds/181153131

gr2m:
okay it’s out now :) https://github.com/hoodiehq/hoodie-client/releases/tag/v7.0.0

package.json
```diff
@@ -13,7 +13,7 @@
   "dependencies": {
     "@hoodie/account": "^2.0.0",
     "@hoodie/admin": "^1.0.1",
-    "@hoodie/client": "^6.0.0",
+    "@hoodie/client": "^7.0.0",
```
gr2m, line 16: when we update `"dependencies"`, we use the `fix:` prefix to force a new release, this helps to deduplicate (sub) dependencies. Only for `"devDependencies"` we use the `chore:` prefix because they don’t get installed by modules that depend on it. 

But for this PR it makes no difference as we already have the `feat:` prefix, so I’m going ahead and merge it as is :)

gr2m (approved):

package.json
sdvg: Thanks, will do so next time :)

Issue #625, opened 2016-10-31
hapi usage: replace options.db with options.PouchDB


gr2m:
refactor options for the hapi plugin usage to be coherent with @hoodie/server, @hoodie/store, @hoodie/account

eg. compare https://github.com/hoodiehq/hoodie#hapi-plugin with https://github.com/hoodiehq/hoodie-server#usage

Taekyoon:
@gr2m I'll take this one!

gr2m:
oh hey @Taekyoon I actually made a camp issue for it: https://github.com/hoodiehq/camp/issues/51 Can you ask Jean there if she still wants to work on it? If yes I am happy to help find something else for you to work on, if not it’s all yorus :)

I’ll close this issue in favor of https://github.com/hoodiehq/camp/issues/51

PR #651, opened 2016-12-08
feat: Deleted statements related "options.db"

Taekyoon:
Deleted statement constructing `options.db` which is not necessary to run.
issue related #625

### DIFF
--- a/cli/parse-options.js
+++ b/cli/parse-options.js
@@ -1,6 +1,8 @@
 module.exports = parseOptions
 
 var log = require('npmlog')
+var PouchDB = require('pouchdb-core')
+var urlParse = require('url').parse
 
 /**
  * Parse options into internal config structure.
@@ -18,14 +20,16 @@ var log = require('npmlog')
  * passe set the `inMemory: true` option. If it’s not set, leveldown is used
  * with the prefix set to `options.data` + 'data' (`.hoodie/data` by default).
  */
+
 function parseOptions (options) {
+  var dbOptions = {}
+
   var config = {
     loglevel: options.loglevel,
     paths: {
       data: options.data,
       public: options.public
     },
-    db: {},
     inMemory: Boolean(options.inMemory)
   }
 
@@ -35,13 +39,30 @@ function parseOptions (options) {
     config.url = options.url
   }
 
-  if (options.dbUrl) {
-    config.db.url = options.dbUrl
-  }
-
   if (options.adminPassword) {
     config.adminPassword = options.adminPassword
   }
+
+  PouchDB.plugin(require('pouchdb-mapreduce'))
+
+  if (options.dbUrl) {
+    if (!urlParse(options.dbUrl).auth) {
+      throw new Error('Authentication details missing from database URL: ' + options.db.url)
+    }
+
+    PouchDB.plugin(require('pouchdb-adapter-http'))
+    dbOptions.prefix = options.dbUrl
+    log.info('config', 'Storing all data in ' + options.dbUrl)
+  } else if (options.inMemory) {
+    PouchDB.plugin(require('pouchdb-adapter-memory'))
+    config.inMemory = true
+    log.info('config', 'Storing all data in memory only')
+  } else {
+    PouchDB.plugin(require('pouchdb-adapter-leveldb'))
+    dbOptions.prefix = config.paths.data + '/data/'
+    log.info('config', 'Storing all data in ' + dbOptions.prefix)
+  }
+  config.PouchDB = PouchDB.defaults(dbOptions)
 
   return config
 }
--- a/server/index.js
+++ b/server/index.js
@@ -3,13 +3,8 @@ module.exports.register.attributes = {
   name: 'hoodie'
 }
 
-var path = require('path')
-var urlParse = require('url').parse
-
 var corsHeaders = require('hapi-cors-headers')
 var hoodieServer = require('@hoodie/server').register
-var log = require('npmlog')
-var PouchDB = require('pouchdb-core')
 var _ = require('lodash')
 
 var registerPlugins = require('./plugins')
@@ -17,46 +12,10 @@ var registerPlugins = require('./plugins')
 function register (server, options, next) {
   options = _.cloneDeep(options)
   _.defaultsDeep(options, {
-    db: {},
     paths: {
       public: 'public'
     }
   })
-
-  // mapreduce is required for `db.query()`
-  PouchDB.plugin(require('pouchdb-mapreduce'))
-
-  if (!options.db.url) {
-    if (options.inMemory) {
-      PouchDB.plugin(require('pouchdb-adapter-memory'))
-      log.info('config', 'Storing all data in memory only')
-    } else {
-      PouchDB.plugin(require('pouchdb-adapter-leveldb'))
-
-      // this is a temporary workaround until we replace options.db with options.PouchDB:
-      // https://github.com/hoodiehq/hoodie/issues/555
-      if (!options.paths.data) {
-        options.paths.data = '.hoodie'
-      }
-
-      options.db.prefix = path.join(options.paths.data, 'data' + path.sep)
-      log.info('config', 'No CouchDB URL provided, falling back to PouchDB')
-      log.info('config', 'Writing PouchDB database files to ' + options.db.prefix)
-    }
-  }
-
-  if (options.db.url) {
-    if (!urlParse(options.db.url).auth) {
-      return next(new Error('Authentication details missing from database URL: ' + options.db.url))
-    }
-
-    PouchDB.plugin(require('pouchdb-adapter-http'))
-    options.db.prefix = options.db.url
-    delete options.db.url
-  }
-
-  options.PouchDB = PouchDB.defaults(options.db)
-  delete options.db
 
   server.ext('onPreResponse', corsHeaders)
 
@@ -64,7 +23,6 @@ function register (server, options, next) {
     if (error) {
       return next(error)
     }
-
     server.register({
       register: hoodieServer,
       options: options
--- a/test/unit/cli/parse-options-test.js
+++ b/test/unit/cli/parse-options-test.js
@@ -29,7 +29,6 @@ test('parse options', function (group) {
     var config = parseOptions({})
 
     t.notOk(config.hasOwnProperty('url'), 'does not set config.url by default')
-    t.notOk(config.db.hasOwnProperty('url'), 'does not set config.db.url by default')
     t.is(config.inMemory, false, 'sets config.inMemory to false by default')
 
     t.end()
@@ -48,9 +47,42 @@ test('parse options', function (group) {
     t.is(config.paths.data, 'options.data', 'uses data option as data path')
     t.is(config.url, 'options.url', 'sets config.url from options.url')
     t.is(config.paths.public, 'options.public', 'uses public option as public path')
-    t.is(config.db.url, 'http://foo:bar@baz.com', 'sets config.db.url from options.dbUrl')
     t.is(config.inMemory, true, 'Sets config.inMemory to boolean equivalent of options.inMemory')
     t.is(config.adminPassword, 'secret', 'Sets config.adminPassword')
+
+    t.end()
+  })
+
+  group.test('dbUrl', function (t) {
+    var config = parseOptions({
+      public: 'public',
+      data: 'data',
+      dbUrl: 'http://foo:bar@baz.com'
+    })
+    var defaults = config.PouchDB('hack', {skip_setup: true}).__opts
+    t.is(defaults.prefix, 'http://foo:bar@baz.com', 'Sets config.db.url')
+
+    t.end()
+  })
+
+  group.test('dbUrl lacking auth', function (t) {
+    t.throws(parseOptions.bind(null, {
+      public: 'public',
+      data: 'data',
+      dbUrl: 'http://baz.com'
+    }))
+
+    t.end()
+  })
+
+  group.test('inMemory', function (t) {
+    var config = parseOptions({
+      public: 'public',
+      data: 'data',
+      inMemory: true
+    })
+
+    t.is(config.inMemory, true, 'Sets config.inMemory to true')
 
     t.end()
   })
--- a/README.md
+++ b/README.md
@@ -71,7 +71,7 @@ option                    | default       | description
 `--address`               | `'127.0.0.1'` | Address to which Hoodie binds
 `--data`                  | `'.hoodie'`   | Data path
 `--dbUrl`                 | –             | If provided, uses external CouchDB. URL has to contain credentials.
-`--adminPassword`         | –             | Password to login to Admin Dashboard. Login is not possible if `adminPassword` option is not set.
+`--adminPassword`         | –             | Password to login to Admin Dashboard. Login is not possible if `adminPassword` option is not set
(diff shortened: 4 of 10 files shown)


gr2m:
Hey @Taekyoon can you rebase your pull request on the latest `master` branch please? Let me know if you need any help with that

Taekyoon:
Hey @gr2m I just rebased my branch and I have no idea with that many commits are on the list. Is it fine to show up these commit signs?
And also I'm not sure that I figure it out correctly. If something mistake on my modifying these codes, can you let me know what I did wrong?

gr2m:
Hey @Taekyoon I think something went wrong with the rebase, I don’t exactly what, but I rebased it now myself and I think the commits look good again?

I’ll do a review as soon as I find some time

Taekyoon:
@gr2m I'll figure it out the problems from my rebase first.

Taekyoon:
@gr2m I tried to rebase again and pushed to my branch, but travis CI reports errors.
The below is the error statements.

```
$ npm test
> hoodie@0.0.0-semantically-released pretest /home/travis/build/hoodiehq/hoodie
> standard
standard: Use JavaScript Standard Style (http://standardjs.com)
  /home/travis/build/hoodiehq/hoodie/server/index.js:6:5: 'path' is assigned a value but never used.
  /home/travis/build/hoodiehq/hoodie/server/index.js:7:5: 'urlParse' is assigned a value but never used.
  /home/travis/build/hoodiehq/hoodie/server/index.js:11:5: 'log' is assigned a value but never used.
npm ERR! Test failed.  See above for more details.
The command "npm test" exited with 1.
Done. Your build exited with 1. 
```
Could you let me know is it a problem from my branch or somewhere else?

gr2m:
The 3 errors you see are coming from standard, a library we use to enforce syntax in all hoodie projects, see https://www.npmjs.com/package/standard

>   /home/travis/build/hoodiehq/hoodie/server/index.js:6:5: 'path' is assigned a value but never used.

You no longer use `path` so you can remove the `var path = require('path')` in line 6.

>   /home/travis/build/hoodiehq/hoodie/server/index.js:7:5: 'urlParse' is assigned a value but never used.

Same with `urlParse` in the line below

>   /home/travis/build/hoodiehq/hoodie/server/index.js:11:5: 'log' is assigned a value but never used.

same with `var log = require('npmlog')` in line 11

server/index.js
```diff
@@ -3,22 +3,15 @@ module.exports.register.attributes = {
   name: 'hoodie'
 }
 
-var path = require('path')
-var urlParse = require('url').parse
-
 var corsHeaders = require('hapi-cors-headers')
 var hoodieServer = require('@hoodie/server').register
-var log = require('npmlog')
 var PouchDB = require('pouchdb-core')
```
gr2m: remove `PouchDB`, it will now be passed as `options.PouchDB` to the `register` function below

server/index.js
```diff
@@ -28,38 +21,6 @@ function register (server, options, next) {
   // mapreduce is required for `db.query()`
   PouchDB.plugin(require('pouchdb-mapreduce'))
```
gr2m: remove this line, it should be moved to `cli/parse-options.js` like most of the rest of the code that you removed :)

Taekyoon: @gr2m I followed your instructions and I just did 'npm test'. But it still has an error related to variable 'config.PouchDB.preferredAdapters'. I have no idea how to deal with this variable which could make an error.
The error messages are the same like this below. 
`not ok TypeError: Cannot read property 'preferredAdapters' of undefined`

Taekyoon: I found it! the problem is about this issue.

https://github.com/pouchdb/pouchdb/issues/5567

gr2m: can you push your current state? I’ll have a look

cli/parse-options.js
```diff
@@ -18,6 +19,9 @@ var log = require('npmlog')
  * passe set the `inMemory: true` option. If it’s not set, leveldown is used
  * with the prefix set to `options.data` + 'data' (`.hoodie/data` by default).
  */
+
+PouchDB.plugin(require('pouchdb-mapreduce'))
+
```
gr2m: remove the `  db: {}` default value defined in line 32. Instead replace it with `  PouchDB: PouchDB`. 

Then above the `if (options.dbUrl) {` block, put in the `PouchDB.plugin(require('pouchdb-mapreduce'))` line, so that the plugin only gets loaded when the server gets started. And then instead of setting `config.db.url = options.dbUrl` we want to set the prefix option on the PouchDB constructor, like so:

```js
config.PouchDB = config.PouchDB.defaults({prefix: options.dbUrl})
```

And then you have to move the logic from the lines 31-58 from `server/index.js` below this block, so that `parseOptions` returns the `config.PouchDB` property will all the right plugins loaded and the right defaults set. 

Makes sense?

Taekyoon: I followed your instructions up to set the prefix option on the PouchDB. But I have no idea with the after mentions. Do you mean that "move codes in 31-58 from `server/index.js` to `cli/parse-options.js` "?

gr2m: yes, exactly! Remember this pull request is about removing `options.db`, and lines 31-58 in `server/index.js` are handling that option right now. So instead we have to move all the handling of it before we pass options to `server/index.js`, and `cli/parse-options.js` is the right place for it

cli/parse-options.js
```diff
@@ -38,8 +38,39 @@ function parseOptions (options) {
...
+    if (options.db.url) {
+      if (!urlParse(options.db.url).auth) {
+        return next(new Error('Authentication details missing from database URL: ' + options.db.url))
+      }
+
+      PouchDB.plugin(require('pouchdb-adapter-http'))
+      options.db.prefix = options.db.url
+      delete options.db.url
+    }
```
gr2m: note that `options.db` does not exist at all. What we want is to gather the default options for our PouchDB constructor and at the end set `config.PouchDB = PouchDB.defaults(dbOptions)`, `dbOptions` will only be used within this `parseOptions` function.

If `options.dbUrl` is set, then load the http adapter plugin and set `dbOptions.prefix`.
If `options.inMemory` is set, load the memory adapter plugin.
Otherwise set `dbOptions.prefix` from `options.paths.data`

If `options.dbUrl` is set, we want `dbOptions.prefix`

Taekyoon: I'm confusing the variable `dbOptions`. Should it be a member of `config`, so that makes 'config.dbOptions'? 
Also, `config.PouchDB = config.PouchDB.defaults({prefix: options.dbUrl})` exists, so I'm not sure that I change the statement from this to `config.PouchDB = PouchDB.defaults(dbOptions)`.

gr2m: define the `dbOptions` variable and use it within the `parseOptions` function only. You can add it right on top

```js
function parseOptions (options) {
  var dbConfig = {}
  // ...
```

I would gather `dbConfig` based on the options (e.g. if `options.dbUrl` is set or `options.inMemory` and then at the at set `config.PouchDB = PouchDB.defaults(dbOptions)`

makes sense?

gr2m:
Hey @Taekyoon I’d like to finish this up soon :) Will you come to Fullstack LA next week? We could pair on it 🙌 💻 🖥 👐

Taekyoon:
I'm so sorry for delaying my work. Because my computer got a problem, I should have recovered and setup it. Now, it's fine to work hoodie again, but it's not an easy job for me to figure out. I'll try out first and I can go to Fullstack next week!!

gr2m:
Cool! And please let me know if you have any questions a long the way. I’m happy to help. It’s always hard to get started, it will get easier with time :) And the change you are working on is not trivial, so don’t feel bad :)

Taekyoon:
@gr2m Does it need to setup `PouchDB.plugin(require('pouchdb-adapter-leveldb'))`, if  `options.inMemory' doesn't set?

gr2m:
>  Does it need to setup PouchDB.plugin(require('pouchdb-adapter-leveldb')), if `options.inMemory' doesn't set?

If neither `options.inMemory` nor `options.dbUrl` are set

cli/parse-options.js
```diff
@@ -34,8 +38,27 @@ function parseOptions (options) {
...
   }
 
+  PouchDB.plugin(require('pouchdb-mapreduce'))
+
   if (options.dbUrl) {
-    config.db.url = options.dbUrl
+    PouchDB.plugin(require('pouchdb-adapter-http'))
+    dbOptions.prefix = options.dbUrl
+
+    if (options.inMemory) {
```
gr2m: this should not be within the `if (options.dbUrl) {` block. If `options.dbUrl` is it means that the data will be written to a CouchDB at  the URL set in `options.dbUrl`, the `options.inMemory` should be ignored at this point.

And the `config.PouchDB = PouchDB.default(dbOptions)` must be run for all cases, so move it out of the `if` blocks.

What you want is something like this:

```js
if (options.dbUrl) {
  PouchDB.plugin(require('pouchdb-adapter-http'))
  dbOptions.prefix = options.dbUrl
} else {
  if (options.inMemory) {
    PouchDB.plugin(require('pouchdb-adapter-memory'))
    log.info('config', 'Storing all data in memory only')
  } else {
    PouchDB.plugin(require('pouchdb-adapter-leveldb'))
    dbOptions.prefix = config.paths.data || '.hoodie'
  }
}

config.PouchDB = PouchDB.default(dbOptions)
```

Does that make sense?

gr2m (changes requested):

cli/index.js
```diff
@@ -26,7 +26,7 @@ function run (callback) {
     var hapiOptions = getHapiOptions(options)
     var server = new Hapi.Server(hapiOptions.server)
     server.connection(hapiOptions.connection)
-
+    debugger
```
gr2m: make sure not to leave `debugger` statements :)

gr2m (changes requested):

gr2m:
when you run `npm test` you should see a few errors

```
  /Users/gregor/Projects/hoodie/hoodie/cli/index.js:29:5: Unexpected 'debugger' statement.
  /Users/gregor/Projects/hoodie/hoodie/cli/parse-options.js:43:10: 'urlParse' is not defined.
  /Users/gregor/Projects/hoodie/hoodie/cli/parse-options.js:44:14: 'next' is not defined.
  /Users/gregor/Projects/hoodie/hoodie/server/index.js:28:1: Trailing spaces not allowed.
```

Are these all clear?

cli/index.js
Taekyoon: Oh I forgot it!! Okay!

cli/parse-options.js
```diff
@@ -34,12 +38,26 @@ function parseOptions (options) {
...
   }
 
+  PouchDB.plugin(require('pouchdb-mapreduce'))
+
   if (options.dbUrl) {
-    config.db.url = options.dbUrl
-  }
-  if (options.inMemory) {
+    if (!urlParse(options.db.url).auth) {
+      return next(new Error('Authentication details missing from database URL: ' + options.db.url))
```
gr2m: Before, this block was within an asynchronous block that required a callback (`next`). But now, the `parseOptions` has no callback option, instead it returns a new value directly. So instead of doing `return next(...)` here, we want to throw an error, like so

```js
throw new Error('Authentication details missing from database URL: ' + options.db.url)
```

makes sense?

Taekyoon: @gr2m I changed it! And I want to ask about test coverage. Can I try to make a code to cover lines in `cli/parse-option.js`?

cli/parse-options.js
```diff
@@ -42,7 +42,7 @@ function parseOptions (options, next) {
 
   if (options.dbUrl) {
     if (!urlParse(options.db.url).auth) {
-      return next(new Error('Authentication details missing from database URL: ' + options.db.url))
+      throw new Error(new Error('Authentication details missing from database URL: ' + options.db.url))
```
gr2m: it’s `throw new Error('Authentication details missing from database URL: ' + options.db.url)`. The `Error` Constructor only accepts a string as argument

gr2m (changes requested):

cli/parse-options.js
Taekyoon: Oh this is my mistake.. Sorry.

gr2m:
okay looking good so far, getting close :)

Now when you run `npm test` you will see a few failing tests. One failing test says "not ok Sets .hoodie as standard paths.data if not given", this is because we removed this logic from [`server/index.js`](https://github.com/hoodiehq/hoodie/pull/651/files#diff-0861d6d6b50d7d695344bf2d86d6e5e6L40). You can remove the entire test `'paths without data-attribute'`, we don’t need it any more. We could add a `test/unit/cli/hoodie-defaults-test.js` file if you like, where we test all the default values including the one for `paths.data` being `'.hoodie'` by default. 

The other error you see "TypeError: Cannot read property 'url' of undefined" is because of the `if () {}` block you moved. Instead of `if (!urlParse(options.db.url).auth) {`, we now check for `if (!urlParse(options.dbUrl).auth) {`. `options.db` no longer exists.

After you fix that, you’ll still see a `TypeError: Cannot read property 'url' of undefined` error, but this time it’s thrown from the `parse-options-test.js` file. We check for `t.is(config.db.url, 'http://foo:bar@baz.com', 'Sets config.db.url')` in one of the tests. Again, `config.db` no longer exists, but here we want to check that the prefix option was set correctly on the `config.PouchDB` constructor. It is not exposed on the constructor, but you can find it out with this workaround

```js
var defaults = new config.PouchDB('hack', {skip_setup: true}).__opts
t.is(defaults.prefix, 'http://foo:bar@baz.com', 'Sets config.db.url')
```

Hope that helps? If anything is unclear, please let me know :)

test/unit/cli/parse-options-test.js
```diff
@@ -58,8 +58,9 @@ test('parse options', function (group) {
       data: 'data',
       dbUrl: 'http://foo:bar@baz.com'
     })
-
-    t.is(config.db.url, 'http://foo:bar@baz.com', 'Sets config.db.url')
+    debugger
```
gr2m: don’t forget to remove the debugger ;)

Taekyoon:
Just fixed it already, but I did't add the test function in `test/unit/cli/hoodie-defaults-test.js`. I'll continue this work till this weekend. Is it fine to keep working this?

gr2m:
yes sure, you are nearly done now :)

Taekyoon:
@gr2m Now I'm trying to make a hoodie-default-test.js. Before I start it, I want to ask that the default value alias `options` value?

Taekyoon:
@gr2m I'm about to make `hoodie-default-test.js`, and I need to know specific information of default values. Which default values are you pointing at? If I notice this values, I can keep making a test file.

gr2m:
Hey @Taekyoon, it’s `hoodie-defaults-test.js` (defaults with an s at the end). The default values are all here: https://github.com/hoodiehq/hoodie/blob/f22ef692d43d5a44f22591575f332d70083d2d90/cli/hoodie-defaults.js.

Taekyoon:
@gr2m Think it's 99.9% done!! :) If it's done, can you recommend me which issues I can handle with?

gr2m:
great work 👏 I can take it from here :) I’ll see what other things there are to be worked on :)

gr2m (approved):
I will merge it locally together with #676 as both are breaking changes

PR #659, opened 2017-01-05
📖👀 http://hoodie.readthedocs.io/

### LINKED ISSUE #109: 📖👀 Let’s use readthedocs.org for Hoodie
The current documentation at http://docs.hood.ie/ is desperately out of date and becoming an ever growing problem and cause of confusion, as it still references the old Hoodie which we no longer maintain. I started with a documentation for the new Hoodie at http://docs.hood.ie/camp/, but it’s not much of a help.

For a long time now I’m a big fan of http://readthedocs.org/, so I used the opportunity and created [a dummy repository](https://github.com/gr2m/hoodie-readthedocs-test/) to see how it would work to use it for Hoodie. Here is the current result: http://hoodie-test.readthedocs.io/en/latest/index.html

My main motivation is to to become more welcoming towards people passionate about documentation, and Read The Docs has a big community. I highly recommend to listen to [Episode 5 of the fantastic RFC podcast](https://changelog.com/rfc/5), which features Eric Holscher, the founder of Read The Docs, [Write The Docs](http://www.writethedocs.org/) and just an overall amazing human being.

Read The Docs has – of course – a great documentation itself: http://docs.readthedocs.io/en/latest/. It is very powerful. Among other things it has great integrations with GitHub, which means each time we would push to `master` branch of the `hoodiehq/hoodie` repository, new documentation is generated. It also supports versioning out-of-the box which works perfectly with our semantic-release setup. And it support translations, including integration with [Transifex](https://www.transifex.com/).

I also like very much that the documentation will become part of the `hoodie` main repository, which will make it simpler to keep both in sync and will us allow to enforce workflows that would not release new feature or breaking versions unless the `docs/` folder is updated, too.

Thinking about downsides, http://readthedocs.org/ is using Python and [Sphinx](http://www.sphinx-doc.org) which I personally am not familiar with myself, but people were super helpful to help me get going, including Eric himself :) Another downside is that the docs use [ReStructuredText](https://en.wikipedia.org/wiki/ReStructuredText) instead of Markdown, but that was the strong recommendation of people from the Write The Docs Community. It’s not that much of a learning curve though, but is much, much more powerful in terms of tooling appreciated by people writing documentation.

Please give this  a thumbs up if you like it, or comment below if you have any concerns.

I would really, really like to go ahead with this, as I see a growing frustration of people new to Hoodie that end up using the old Hoodie version instead of the new one.

ping @hoodiehq/maintainers

gr2m:
closes hoodiehq/discussion#109

Once merged:

- [ ] setup DNS to make http://docs.hood.ie/ point to hoodie.readthedocs.io, see http://docs.readthedocs.io/en/latest/alternate_domains.html


minrwhite:
Since hoodiehq/hoodie-client#110 has now been merged, should this documentation reference 'localstorage' or 'IndexedDB'? Or in this documentation does 'localstorage' just mean 'some form of persistent storage local to the client', in which case I'd probably want clarification by removing the camelcased (`localStorage`) instances of this in the documentation, and preferably to have the word split back into 'local storage', and an extra explanatory section somewhere.

gr2m:
Thanks @benwhite-deltas, I removed the mentions of `localStorage`, good catch! I did not correct the image yet though, I will make a follow-up issue for it.

I’m sure there are other problems in there, this is mostly a copy of what we currently have at http://docs.hood.ie/camp/. But now it should be simpler to contribute, as each page has an "Edit on GitHub" button.

We got all the `+1`s on https://github.com/hoodiehq/discussion/issues/10 and no objections, so I will go ahead and merge it.

PR #664, opened 2017-01-08
docs(quickstart): OS -> operating systems

gr2m:
Avoiding abbreviations is the Hoodie way™️

catch me if you can


minrwhite (approved):
Clarity FTW :wink:

Issue #660, opened 2017-01-06
Add documentation for docs contributors / maintainers


gr2m:
follow up for https://github.com/hoodiehq/discussion/issues/109#issuecomment-270891487

Every page on http://hoodie.readthedocs.io/ has an "Edit on GitHub" button on the top right corner, so making small changes should be no problem. But for bigger changes, we should document on how to setup / build the documentation locally. 

Read The Docs is using [Sphinx](http://sphinx-doc.org/) which is a Python project. This Question might be relevant, too http://docs.readthedocs.io/en/latest/faq.html#i-want-to-use-the-read-the-docs-theme-locally

gr2m:
How about we add a `docs/README.md` file that will be shown on GitHub when opening the [docs folder](https://github.com/hoodiehq/hoodie/tree/master/docs), with information on what this folder is, how things work, how to contribute and how to setup it up locally?

ThomasWaldmann:
Not sure if this helps, but see yourself:

See there about how we build local / on rtfd: https://github.com/borgbackup/borg/tree/master/docs

- we have slightly customized the readthedocs theme (logo, colours)
- see also our docs, development section about how to build the docs

fozy81:
+ for to add documentation for docs contributors / maintainers

We could get a little more testing functionality out of Sphinx beyond flagging dead url links. For example by adding a [spell checking module](https://doughellmann.com/blog/2011/05/26/creating-a-spelling-checker-for-restructuredtext-documents/) too? - I know my spelling is bad! Perhaps travis CI can run the checks on the pull request? 

Beyond that, there's some interesting technical document testing ideas mentioned in this blog post https://www.mapbox.com/blog/retext-mapbox-standard/

gr2m:
great idea! And yes, I’m sure we can add more checks to our travis builds :)

fozy81:
I like the idea to treat 'docs-as-code'. There's a interesting plain [text linter](https://github.com/textlint/textlint) which can test for all types of typos and errors. It also works with [rst](https://github.com/jimo1001/textlint-plugin-rst). 

Open Stack docs has a nice [style guide](http://docs.openstack.org/contributor-guide/writing-style/general-writing-guidelines.html) for technical docs. Maybe of interest, to people new to writing docs (like me!).

PR #665, opened 2017-01-08
fix 404s

quangv:
For the API 404s, not sure if I did it right.

for example if you go here: https://github.com/quangv/hoodie/blob/master/docs/api/index.rst it's still 404, it'll be nice if the links work on both http://docs.hood.ie/ && github.

### DIFF
--- a/public/index.html
+++ b/public/index.html
@@ -24,7 +24,7 @@
         <li> Setup your asset folders and create your .css and .js files.</li>
         <li> You’ll need to restart the Hoodie server so it servers assets from your app’s <code>public</code> folder</li>
       </ol>
-      <p>If you want to learn more, check out the <a href="http://docs.hood.ie/camp/"> documentation (work-in-progress).</a></p>
+      <p>If you want to learn more, check out the <a href="http://docs.hood.ie"> documentation (work-in-progress).</a></p>
       <p>Happy coding! 🐶</p>
     </div>
 


gr2m:
I was trying to setup [sphinx](http://www.sphinx-doc.org) so I can build and test the documentation locally. Right now the only way to test this is to merge and 🙏 and hope the result on docs.hood.ie is as desired :) 

I don’t know any python and I keep running into trouble have to give up on it due to a lack of time. This is somewhat related to https://github.com/hoodiehq/hoodie/issues/660. 

We can reach out to see if someone from the Python / Sphinx / Read The Docs community can help us with instructions to build the docs locally, then we can test your pull request locally and make sure it doesn’t break the build docs on docs.hood.ie

fozy81:
Hi Gregor, I googled some of the error codes you had with building docs locally - didn't find a remedy, but could be an issue on the Sphinx side. On linux, I get a lot of warnings but still works okay. Maybe worth building on a different osx machine to see if specific to your machine or osx?

If error with Sphinx, could post an issue here https://github.com/sphinx-doc/sphinx/issues ? 

We could hard-code the links to http://docs.hood.ie/... but not sure that's ideal.  Anyway, currently all links mentioned in this PR have been 'fixed' - at least on Read the Docs. The only remaining  issue is the link on public/index.html. 

All the best,

Tim

gr2m (approved):

gr2m:
thanks @quangv & @fozy81! I’ve rebased and the PR and squashed the commits to adjust the commit message 👍

PR #676, opened 2017-01-23
drop LevelDB dependency

### LINKED ISSUE #624: Drop levelDB dependency
Instead of `pouchdb-adapter-leveldb` use `pouchdb-adapter-fs` by default. Make the pouchdb adapter configurable with a new option and document it in the release notes so existing apps can easily be migrated by adding `pouchdb-adapter-leveldb` as dependency and by configuring it as adapter

gr2m:
closes #624. I tested it with the acceptance tests in `hoodie-app-tracker` and manual tests, works all good.

This will speed up our installation time and make Hoodie even simpler to install as we drop the `node-gyp` dependency and the `leveldb` postinstall build step, which currently breaks deployment to now.sh, for example


gr2m:
I’ll merge this together with #651 as both are breaking changes

PR #678, opened 2017-01-23
assure data folders

gr2m:
We currently assure the folders in `@hoodie/server`, but the correct place for it is in `hoodie/cli`. Part of hoodiehq/hoodie-server#532

### DIFF
--- a/cli/assure-folders.js
+++ b/cli/assure-folders.js
@@ -1,0 +1,22 @@
+module.exports = assureFolders
+
+var path = require('path')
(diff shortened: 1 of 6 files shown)


gr2m:
I’ve tested it locally extensively. This is fixing deployment to know which several folks are waiting for, so going ahead and merge it. But if you find anything let me know and I’ll create follow up issues / PRs

sdvg (commented):
Looks good to me :)

PR #690, opened 2017-02-19
initial commit for 3rd-party-plugin

jicksy:
Hi @gr2m 

My pull request. 
 
I tried assigning the value of hoodie.plugins array to thirdPartyPlugins, is this the correct way to do it? It gives me error hoodie.plugins module is not present in my locale.


server/plugins/client/bundle.js
```diff
@@ -31,6 +31,14 @@ function bundleClient (hoodieClientPath, bundleTargetPath, config, callback) {
   var plugins = [
     path.resolve('hoodie/client')
   ].filter(checkModule)
+
+  var thirdPartyPlugins = require('hoodie.plugins')
```
gr2m: `require('hoodie.plugins')` looks for a module with the name `hoodie.plugins` that would need to be listed in hoodie’s [package.json dev/dependencies](https://github.com/hoodiehq/hoodie/blob/b2f16b827148ec8478571f42572292d635abc847/package.json#L13-L49).

Instead you want to look into `config.plugins`. We don’t currently have a `plugins` option, so you would need to add that. Maybe not for the CLI usage, but for the [hapi usage](https://github.com/hoodiehq/hoodie#hapi-plugin). The default value should be `[]` and you can set it in https://github.com/hoodiehq/hoodie/blob/master/cli/hoodie-defaults.js

To test all this, I suggest you create a new Hoodie app as described in http://docs.hood.ie/en/latest/guides/quickstart.html, and then link your local `hoodie` repository into your app, with `npm link /path/to/hoodie`. That way you can restart your app and the changes you do to your local `hoodie` repository will directly apply.

In your app’s `package.json`, add `"hoodie": {"plugins": ["hoodie-example-plugin"]}`, and install `hoodie-example-plugin` as a dependency.

Now if you restart your app, `config.plugins` should be `['hoodie-example-plugin']`, so you can do `var thirdPartyPlugins = config.plugins

I know it’s quite a lot, I hope it’s not too overwhelming?

server/plugins/client/bundle.js
```diff
@@ -31,6 +31,14 @@ function bundleClient (hoodieClientPath, bundleTargetPath, config, callback) {
...
+  $.each(thirdPartyPlugins, function(){
+    concat(
+    path.resolve('hoodie/client').filter(checkModule)
+    )
+  })
```
gr2m: `$.each` is a jQuery function, and we don’t use jQuery in the Hoodie server code. What you can do instead is this:

```js
thirdPartyPlugins.foreach(function (pluginName) {
  var pluginPath = pluginName + '/hoodie/client'
  if (checkModule(pluginPath)) {
    plugins.push(require.resolve(pluginPath))
  }
})
```

[require.resolve](https://nodejs.org/api/globals.html#globals_require_resolve) returns the full path to the file we want to bundle into the client.

server/plugins/index.js
```diff
@@ -43,6 +43,20 @@ function registerPlugins (server, config, callback) {
...
+    concat(
+    path.resolve('hoodie/server').filter(checkModule)
+    )
+    .map(function (register) {
+      return {
+        options: options,
+        register : require(register)
+      }
+    })
+  })
```
gr2m, line 58: see my comments above, it is similar with the server-side plugins. You can find out if a server extension exists in a 3rd party plugin using `checkModule('name-of-the-npm-package/hoodie/server')`. If it returns true, add it to the `thirdPartyPlugins` array and then pass it all together to `server.register` with `hapiPlugins.concat(thirdPartyPlugins).concat(localPlugins)`

gr2m (changes requested):

jicksy:
I have edited package.json and bundle.js files. Index.js is pending.

package.json
```diff
@@ -67,5 +67,6 @@
     "pretest": "standard",
     "test": "nyc tap --no-cov './test/{unit,integration}/**/*-test.js'",
     "postinstall": "node ./bin/setup.js"
-  }
+  },
+  "hoodie": {"plugins": []}
```
gr2m, line 71: you add that to your app which has `hoodie` as its dependency, we don’t need to add it to `hoodie` iteslf

server/plugins/client/bundle.js
```diff
@@ -32,11 +32,13 @@ function bundleClient (hoodieClientPath, bundleTargetPath, config, callback) {
...
-    concat(
-    path.resolve('hoodie/client').filter(checkModule)
-    )
+  var thirdPartyPlugins = config.plugins
+
+  thirdPartyPlugins.foreach(function (pluginName) {
+  var pluginPath = pluginName + '/hoodie/client'
+  if (checkModule(pluginPath)) {
+    plugins.push(require.resolve(pluginPath))
+    }
```
gr2m, line 41: could you please fix the indentation? Add two spaces to the beginning of line 38, 39 and 40

cli/hoodie-defaults.js
```diff
@@ -11,6 +11,7 @@ function getHoodieDefaults () {
     inMemory: false,
     loglevel: 'warn',
     url: undefined,
-    adminPassword: undefined
+    adminPassword: undefined,
+    hoodie.plugins=[]
```
gr2m, line 15: it’s a `:`, not an `=`

gr2m:
To see if you current changes broke anything unintended, you can always run `npm test`. If you see an error it means you probably have broken something by accident. It can help you to find the problems early :) but don’t worry, you don’t have to, I’ll walk you trough. And please let me know if there is anything that does not make sense

jicksy:
Hi @gr2m 
I'm away for one week on something personal. If anyone wants to take this up during this week, feel free to re-assign.

PR #693, opened 2017-02-22
don’t stop the logging

gr2m:
I run into a problem where logging would simply stop working after a moment. I don’t understand exactly why, but this change fixes it

### DIFF
--- a/server/plugins/good-hoodie.js
+++ b/server/plugins/good-hoodie.js
@@ -51,7 +51,7 @@ function transform (data, enc, next) {
   }
 
   log.silly(data.event, data)
-  next(null, data)
+  next(null)
 }
 
 function findLogLevel (levels, tags) {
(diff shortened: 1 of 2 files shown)


server/plugins/good-hoodie.js
```diff
@@ -51,7 +51,7 @@ function transform (data, enc, next) {
   }
 
   log.silly(data.event, data)
-  next(null, data)
+  next(null)
```
gr2m, line 54: from what I know the 2nd argument should always be a string. The other good plugins `good-console` passes the string it wants to be logged: https://github.com/hapijs/good-console/blob/c4fb48f2bb50b0d346d8814d01c0316f40f002f7/lib/index.js#L166

We log ourselves with npmlog, so we don’t need to pass anything. It probably is not the right way to use good, but at least with that change it won’t simply stop logging randomly

PR #694, opened 2017-02-22
feat: updated developer documents and lint tests

fozy81:

PR #705, opened 2017-02-28
doc: get readthedocs to build

fozy81:
Appears deleting the conf.py lets the readthedocs build. Not sure why, are there two conf.py? something to do with gitignore?


fozy81:
fixes: https://github.com/hoodiehq/hoodie/pull/694

gr2m:
@fozy81 Great, thanks :)

Two things unrelated to this PR:

- for documentation fixes, we use the `docs: ` prefix for our commit messages. A `fix: ` implies that we fixed a bug in the code and it would automatically release a new version. But in that case we just want to update the documentation which happens for each commit on master automatically.
- you can create branches on the main repo (https://github.com/hoodiehq/hoodie) instead of on your fork. It makes things a bit simpler for us. For example I needed to rebase your PR on latest master, and it’s simpler for me to do so with a branch on the main repo compared to a branch on your fork, it needs less git commands :)

gr2m (approved):

fozy81:
@gr2m Thanks for your review, really appreciated. I'll branch next time and use the doc: commit message for updates to docs. :+1: 

There's a few more changes I'd like to make, as building the docs locally now needs looking at. But I think the main thing was to fix the readthedocs pages - and they are correct now

PR #707, opened 2017-02-28
Db url special chars fix | Undefined options.db.rul fix | READY FOR MERGE

### LINKED ISSUE #668: Config hoodie/dbUrl not accepted when password contains some special chars
Let's say dbUrl is set to https://USERNAME:PASSWORD@ACCOUNTNAME.cloudant.com/

If PASSWORD contains is too complicated (contains slashes, semicolon - chars have to be discovered as I did not test them specifically!) the app cannot start and the following error is thrown:

Error: Authentication details missing from database URL

### LINKED ISSUE #699: options.db.url is not specified
Where does options.db.url is being initialized and set, in order to be used at https://github.com/hoodiehq/hoodie/blob/master/cli/parse-options.js#L51 ?

Is that some sort of left-over code from previous fixes? Have I found a bug?

Manosmer:
Fixes #668. Also, fixes #699, because the test case for lacking of auth details in dbUrl is covered in function createAuthDbUrl and if(parseUrl(options.dbUrl).auth) case (which contained options.db.url) is never reachable and was removed.


bin/start.js
```diff
@@ -1,11 +1,15 @@
...
+var log = require('npmlog')
+log.level = 'warn'
+
 var cli = require('../cli')
 
 var emoji = require('node-emoji')
 
 cli(function (error, server) {
   if (error) {
+    log.warn('DB url config', 'If you are using --dbUrl, --dbUrlUsername, --dbUrlPassword, you should provide your authentication details without URL encoding')
```
gr2m: There can be many errors that can returned by the `cli` callback, we cannot assume that is related to the DB url config. I think this is the wrong place to log this

gr2m: There can be many errors that can be returned by the cli callback, we cannot assume that is related to the DB url config. I think this is the wrong place to log this

cli/options.js
```diff
@@ -67,7 +67,17 @@ function getCliOptions (projectPath) {
...
+      },
+      dbUrlPassword: {
+        type: 'string',
+        default: defaults.dbUrlPassword,
+        describe: 'Provides the password (if required) for auth with the db at dbUrl (requires dbUrl and/or dbUrlUsername)'
+      },
+      dbUrlUsername: {
+        type: 'string',
+        default: defaults.dbUrlUsername,
+        describe: 'Provides the username (if required) for auth with the db at dbUrl (requires dbUrl and dbUrlPassword)'
```
gr2m: note that username & password are always required by Hoodie

cli/parse-options.js
```diff
@@ -20,9 +19,89 @@ var urlParse = require('url').parse
  * PouchDB is being used. A shortcut to set PouchDB’s adapter to memdown is to
  * passe set the `inMemory: true` option. If it’s not set, leveldown is used
  * with the prefix set to `options.data` + 'data' (`.hoodie/data` by default).
+ *
+ * @function createAuthDbUrl accepts options.dbUrlUsername, options.dbUrlPassword, options.dbUrl, @returns encoded dbUrl
```
gr2m: we don’t use these `@function` or `@returns` tags anywhere yet. I don’t mind starting to do them, we were looking into tools to generate API docs out of comments. But I’d coordinate the two: the `@name` tags with a tool that can parse them and generate docs out of them.

I suggest we make a follow up issue for this, and for the time being do not use the tags, as their usage implies that we have some API docs generator?

cli/parse-options.js
```diff
@@ -20,9 +19,89 @@ var urlParse = require('url').parse
...
+   * @param dbUsername used for parsing options.dbUrlUsername
+   * @param dbPassword used for parsing options.dbUrlPassword
+   * @param dbUrl used for parsing options.dbUrl
+   *
+   * @var dbUrlParts contains different parts of dbUrl, helping in modifing username and password
+   *
+   * @returns dbUrl based on the parameters
+   * @throws Error if dbUrl is unparsable or authDetails are missing
+   */
+  function createAuthDbUrl (dbUsername, dbPassword, dbUrl) {
```
gr2m: we try to avoid defining a function within another function. This one is quite big, can you move it into `cli/utils/create-auth-db-url.js`?

test/unit/cli/parse-options-test.js
```diff
@@ -68,6 +72,57 @@ test('parse options', function (group) {
...
+    var config = parseOptions({
+      public: 'public',
+      data: 'data',
+      dbUrl: 'https://https://foo@baz.com'
+    })
+    var defaults = config.PouchDB('hack', {skip_setup: true}).__opts
+    t.is(defaults.prefix, 'https://https:%2F%2Ffoo@baz.com', 'Sets config.db.url')
+
+    t.end()
+  })
```
gr2m, line 100: does that test make sense? Is the double `https://https` an accident?

test/unit/cli/parse-options-test.js
```diff
@@ -68,6 +72,57 @@ test('parse options', function (group) {
     t.end()
   })
 
+  group.test('dbUrl', function (t) {
```
gr2m: can you make the test labels more descriptive? Like `dbUrl with encrypted url encoded username` etc?

test/unit/cli/parse-options-test.js
```diff
@@ -68,6 +72,57 @@ test('parse options', function (group) {
...
+    var config = parseOptions({
+      public: 'public',
+      data: 'data',
+      dbUrl: 'http://dasds@dasdsa.com:@dsadas@dasdas@@baz.com'
+    })
+    var defaults = config.PouchDB('hack', {skip_setup: true}).__opts
+    t.is(defaults.prefix, 'http://dasds%40dasdsa.com:%40dsadas%40dasdas%40@baz.com', 'Sets config.db.url')
+
+    t.end()
+  })
```
gr2m: can you add a test that has both auth in the URL but also passes `dbUrlUsername` and `dbUrlPassword` so we can define how Hoodie is supposed to behave in such a case?

Manosmer:
Hello @gr2m 😄. Did you delete your comment? Because I can't find it anymore. Also, I am adding myself some new test cases in parse-options-test.js. I am not finished with that one yet. So, the coverage will be raised again to 99%. The PR was made in order to make you able to review my code and make a request or change before the merge.

gr2m (changes requested):

gr2m:
@Litarhis so sorry I removed the comment to make it part of my review and then forgot to actually submit my review lol

bin/start.js
Manosmer: Thought so too. Do you have any opinion about where to put that warning?

cli/parse-options.js
Manosmer: Got it!

test/unit/cli/parse-options-test.js
Manosmer: I was just playing with test cases for the moment. I will leave them as a final job. I would like to finish with the code at first.

test/unit/cli/parse-options-test.js
Manosmer: For my code, arguments `--dbUrlUsername` and `--dbUrlPassword` have a higher priority than auth details provided in `--dbUrl` (I think it's more logically correct), thus they will override the auth details included in `--dbUrl`. I am going to rewrite the tests, add more of them and describe each one of them clearly

bin/start.js
gr2m: My guess is that it should go directly to the place where we parse the URL. If the parsed URL has no authentication, then we should log an explanation with `log.error`

Manosmer: The warning is present to inform the users in case they provided auth details already encoded in URL format. In that case, if the user wanted to pass `http://john@doe.com:pass@example.com`, but in an already encoded format, namely `http://john%40doe.com:pass@example.com`, the symbol `%` in the username would be encoded again and the URL would be: `http://john%4040doe.com:pass@example.com`. This is, obviously, wrong.

gr2m: Agree, but it’s the wrong place to show this warning. It should go to the place the parsing is happening

test/unit/cli/parse-options-test.js
gr2m: I agree with the priority. I would still write a test for it so we don’t accidentally break behavior in future. It would also be a nice-to-have to inform the user that a username/password is set in both `dbUrl` and `dbUrlUsername` and that the former one is being ignored

test/unit/cli/parse-options-test.js
Manosmer: No it's not 😛. It's in case the user's username starts with https:// 😉

test/unit/cli/parse-options-test.js
Manosmer: I think in my code I warn the user for the override. Also, I will split the code and create a new file (cli/utils/create-auth-db-url.js) just like you proposed. So it would be required to transfer some of the current tests to another `*-test.js` file

Manosmer:
Hello @gr2m. I would like to inform you that I will be off until Sunday due to some difficulties. I will start fixing this PR from Monday. Sorry for any inconvenience.

gr2m:
no worries, thanks for keeping us posted

Manosmer:
At last! First of all, good day @gr2m! Sorry for my absence. I finally found some time to focus on this project and I think that it is finished and working fine. I have only one issue I cannot resolve here: https://github.com/hoodiehq/hoodie/blob/dbUrl-specialChars-fix/test/unit/cli/parse-options-test.js#L195 . Test gives back `not ok path must be a string`. Can you help me with that? I think after that we will finish with this one.

gr2m:
awesome, thanks! 

> I have only one issue I cannot resolve here: https://github.com/hoodiehq/hoodie/blob/dbUrl-specialChars-fix/test/unit/cli/parse-options-test.js#L195 . Test gives back not ok path must be a string

You need to add `dbAdapter: 'pouchdb-adapter-fs'` to the options passed to `parseOptions`, otherwise you try to require(undefined) in https://github.com/hoodiehq/hoodie/blob/e9fd5e143c9ff01f6bbe43e34428c65fa8feaeb4/cli/parse-options.js#L63

You can then test if the adapter is `fs` as expected with

```js
var config = parseOptions({ ... })
t.is(config.PouchDB.preferredAdapters[0], 'fs')
````

**Comment by gr2m** [comment]
by the way, check out debugging Node with Chrome Dev Tools: https://nodejs.org/api/debugger.html#debugger_v8_inspector_integration_for_node_js

Add a `debugger` on top of the failing test, then run the tests with `node --inspect --debug-brk test/unit/cli/parse-options-test.js`. It’s super awesome ✨

**Comment by Manosmer** [comment]
Awesome, @gr2m! Thank you

File: .gitignore [hunk 8]
```diff
@@ -5,3 +5,4 @@ node_modules
 .hoodie
 .nyc_output
 docs/_build/
+.vscode
```
gr2m: can you leave this out of the `.gitignore`? You can add it to a `.gitignore` in folder above the repository. I have things like `node_modules` in `~/.gitignore`

Other than that this is good to merge :)

gr2m (changes requested):

.gitignore
Manosmer: Allright 😄

cli/options.js
gr2m: can you please remove the `(if required)` in the descriptions above?

Manosmer:
@gr2m I think it's ready to go now 😛

gr2m (approved):
well done 👏

gr2m:
ugh totally forgot to clean up commits

Litarhis:
Can the corresponding branch be deleted or is it still useful?

gr2m:
you can delete them once they are merged :)

Manosmer:
Because you said something about cleaning up the commits... I didn't catch that..:( also sorry for restoring and deleting... I am from my mobile's small screen

gr2m:
no worries, fixed it

PR #711, opened 2017-03-02
3rd party plugins

### LINKED ISSUE #104: Enable 3rd party plugins
This is something we have been working towards for the past year, and finally the pieces are all in place in order to allow to extend your Hoodie app easily with plugins provided by npm packages.

We created a [dummy plugin](https://github.com/hoodiehq/hoodie-plugin-example) with extended information on how things should work. The two blockers we mention in there are now done.

### :dart: The goal
The goal is to get the plugin example to work. When I add it to my app, set `options.plugins = ["hoodie-plugin-example"]` and start my app, I should have `hoodie.send.email()` method in the client which sends the `POST /hoodie/email/api/send` which then gets handled by the route handler defined in the plugin that actually sends an email. 

I’m sure there are some things we didn’t think of yet, pleas reach out if you have any questions or run into any blocker. Enjoy :)

labaikie:
closes https://github.com/hoodiehq/camp/issues/104

### Update

This PR is ready for review, see https://github.com/hoodiehq/hoodie/pull/711#issuecomment-304530166


gr2m:
I think the next steps are these

- figure out how to load the client bit of the 3rd party plugin
- figure out to serve the contents of `/hoodie/public` folder (if existent) at `<hoodie url>/hoodie/<plugin name>/*`
- make sure configuration is passed to the plugin (see https://github.com/hoodiehq/hoodie-plugin-example/blob/36c80c1e6dfb02dfbed9ecd27731b83a4a9a88fa/hoodie/server.js#L18)
- write tests and make them pass :)

We can use this pull requests to discuss, let us know if you have any questions :)

labaikie:
Thanks! I do have a lot on my plate until Wed next week but I hope to take another look at it over this weekend and let you know if I have any further questions!

labaikie:
It looks like the server plugin integration is not yet complete 😞 ... 
In running a local test hoodie app with the hoodie-plugin-example included as a plugin (and after temporarily replacing `options.email.transporter` with the an email smtp string in plugin [here](https://github.com/hoodiehq/hoodie-plugin-example/blob/36c80c1e6dfb02dfbed9ecd27731b83a4a9a88fa/hoodie/server.js#L18)), the test app gets terminated with the below error.

```
Error: Plugin hoodie-email already registered in: http://localhost:8080
<     at Object.exports.assert (../hoodie/node_modules/hapi/node_modules/hoek/lib/index.js:736:11)
<     at each (../hoodie/node_modules/hapi/lib/plugin.js:288:26)
<     at iterate (../hoodie/node_modules/hapi/node_modules/items/lib/index.js:36:13)
<     at done (../hoodie/node_modules/hapi/node_modules/items/lib/index.js:28:25)
<     at Object.module.exports.register (../testapp/hoodie/server.js:6:3)
<     at Object.target [as register] (../hoodie/node_modules/hapi/node_modules/joi/lib/object.js:77:34)
<     at each (../hoodie/node_modules/hapi/lib/plugin.js:318:14)
<     at iterate (../hoodie/node_modules/hapi/node_modules/items/lib/index.js:36:13)
<     at done (../hoodie/node_modules/hapi/node_modules/items/lib/index.js:28:25)
<     at Object.register (../hoodie/server/plugins/public.js:94:10)
```

Would you by any chance have any pointers? Thank you!

gr2m:
This looks like an error you get when you try to register a plugin with the same name twice with hapi.

In your test app, in the `/hoodie/server.js` file, do you register a plugin that is called `hoodie-email`, just like the plugin is called in `hoodie-plugin-example` module? That’s probably the problem :) You can rename the plugin name in `/hoodie/server.js` in your testapp, or you can remove the `hoodie/` folder altogether, it was just to show how loading the plugins should work. Now that we load the plugin from  `hoodie-example-plugin/plugin/server` we are good and don’t need it any longer for testing

labaikie:
@gr2m hey so I tried removing & changing the name of the module being registered at my testapp's `/hoodie/server.js` but no luck in getting the hoodie server to not terminate. :( I'm trying to debug through the processes but having a hard time - are you by any chance available for some questions?

labaikie:
hi @gr2m - so I'm trying to load the plugin's `public` contents and I'm having trouble specifying the plugin paths in `hoodie/server/public.js`. Please see the snippet below and let me know if you would have any pointers. I've tried a few variations, haha.

```
const routes = [{
    method: 'GET',
    path: '/{p*}',
    handler: {
      directory: {
        path: publicFolder,
        listing: false,
        index: true
      }
    }
  }, {
    method: 'GET',
    path: '/hoodie/{p*}',
    handler: {
      directory: {
        path: hoodiePublicPath,
        listing: false,
        index: true
      }
    }
  }, {
    method: 'GET',
    path: '/hoodie/admin/{p*}',
    handler: {
      directory: {
        path: adminPublicPath,
        listing: false,
        index: true
      }
    }
  }, {
    method: 'GET',
    path: '/hoodie/info.json',
    handler: function (request, reply) {
      reply({
        hoodie: true,
        name: options.config.name,
        version: hoodieVersion
      })
    }
  }]

  // add plugin routes
  plugins.forEach(i => routes.push({
    method: 'GET',
    path: `/hoodie/${i}/{p*}`,
    handler: {
      directory: {
        path: pathJoin(require.resolve(`@hoodie/${i}`), 'hoodie', 'public'),
        listing: false,
        index: true,
      }
    }
  }))
```

gr2m:
yeah that’s a tricky one: finding the path to a node_modules package. I’m sure there is a package for that :D 

What I usually do is this

```js
var path = require('path')
var packageName = 'my-npm-package'
var packagePath = path.dirname(require.resolve('hoodie/package.json'))
var pathToPublicFolder = path.join(packagePath, 'hoodie', 'public')
```

The trick is that each package must always have a  `package.json` file, and require supports `*.json` files.

Note that you should check if the folder exists before defining the route.

Hope that helps :)

labaikie:
@gr2m thanks! I got it - so it's `${pacakgeName}/package.json`!

labaikie:
@gr2m where do you think is best for the users of hoodie app to specify the plugin configuration? For the example email plugin, the smtp transporter, namely? I was wondering if it would be the app's `package.json`? Thanks!

gr2m:
yes I would put the config into `package.json`. We can add more keys to the "hoodie" property, like

```js
{
  ...
  "hoodie": {
    "public": "dist/",
    "plugins": ["my-plugin"],
    "my-config": {...}
   }
}
```

Something like this? We need to discuss this a bit more, but for now that’s good enough. Does that work for ya?

labaikie:
@gr2m, thanks so much for all your support. I've pushed up the code to include

a. server plugin load
b. client plugin load
c. public content load
d. option to pass plugin configuration

Would you be able to help me review? Thank you!

gr2m:
This is great work, @labaikie 👏 

There are a few things left to do, and I’m happy to help you finish it up or do it for you.

1. We don’t yet use any of the es2015 syntax like `const` and destructing. We will at some point move to it, but have to plan it a bit, so for now it would be good if you would keep it with `var`
2. When loading the example plugin, the route for the public folder (`GET /hoodie/hoodie-plugin-example/*`) seems to override the one that is registered in the plugin (`POST /hoodie/hoodie-plugin-example/api/send`). we probably might need to change order of things to fix that.
3. I would like to introduce a new option `"hoodie": {"name": "my-custom-name"}` so that the example plugin for example could set it to `"email"`, which we then could use for the routes, so we would have `/hoodie/email/*` instead of `/hoodie/hoodie-plugin-example/*`. What do you think?
4. We will need to test it all, but I can definitely help with that :)

I’d like to move this forward, today / tomorrow. Let me know if you want to give it a go yourself

janl:
Great progress here!

I started on porting to ES5: https://github.com/hoodiehq/hoodie/pull/745 and while doing so I found one tiny issue mentioned in #745

gr2m:
The current work-in-progress implementation has a configuration problem: all plugin developers can decide to add keys for their configuration to the `"hoodie"` object in `package.json`. If one day we decide to add a new key that hoodie core is using, it might conflict with 3rd party plugins out there.

This is something we should discuss with the wider community in future, but for now @janl and I think the best ways is to put all app/plugin-specific configuration into an extra key, that we suggest to call `"app"`.

So instead of 

```json
{
  ...
  "hoodie": {
    "plugins": ["hoodie-plugin-example"],
    "example": {"foo": "bar"}
  }
}
```

it would now be 

```json
{
  ...
  "hoodie": {
    "plugins": ["hoodie-plugin-example"],
    "app": {"example": {"foo": "bar"}}
  }
}
```

As a reminder: the reason why we don call `"app"` something like `"pluginOptions"` or make `"plugins"` an object where keys are the plugin package names and the values are the options is that we want hoodie plugins to be valid hoodie apps and the other way around :)

gr2m:
@fozy81 Hey Tim, I run into a problem with the link checker. The PR is adding a link to the README which links to a part of the docs that is only created with this PR itself, so the link fails. For now I whitelisted docs.hood.ie in `.travis.yml`, but maybe you have a better idea?

Besides that, the link checker fails for URLs like https://pouchdb.com/2016/06/06/introducing-pouchdb-custom-builds.html, which works for me™️, any idea what that could be? https://travis-ci.org/hoodiehq/hoodie/builds/236912318#L1029

gr2m:
I don’t get the error "Error: Cannot find module 'pouchdb-browser' from '/tmp/tmp.ifWeTYm9KO' " https://travis-ci.org/hoodiehq/hoodie/builds/236912318#L611 because above it shows that it was installed successfully ... https://travis-ci.org/hoodiehq/hoodie/builds/236912318#L571 ugh

I think this is unrelated, I made a follow up issue here: https://github.com/hoodiehq/hoodie/issues/749

I’ll enable the docs checks again so we can get this fixed, and then it’s ready to merge

gr2m:
@hoodiehq/maintainers this is ready for review 👀 

In order to test the app, create a new app as described in http://docs.hood.ie/en/latest/guides/quickstart.html. Instead of installing from npm, install hoodie from GitHub

```
npm install --save hoodiehq/hoodie#camp/104/3rd-party-plugins
```

Then add an app-specific plugin as described in https://github.com/hoodiehq/hoodie/blob/camp/104/3rd-party-plugins/docs/guides/plugins.rst and then install the [hello world plugin](https://github.com/hoodiehq/hoodie-plugin-hello-world) (make sure not to forget to set `hoodie.plugins` to `["@hoodie/plugin-hello-world"]` in your package.json as described in the README!)

When you start the app now, it should load your app specific plugin (if you haven’t changed the example code, you should see "Hello, world!" at `http://localhost:8080/hoodie/<your app name>/api`) and when the hello world plugin works it should show you a form at http://localhost:8080/hoodie/hello-world

fozy81:
@gr2m Hi, a quick fix to reference documents within repo is do something like this: <CONTRIBUTING.html> or in this particular case <hapi-plugin.html> 
Here's an example .rst: https://raw.githubusercontent.com/hoodiehq/hoodie/master/docs/developers/CODING_STYLE.rst

I think this should solve the issue.

I'll need to look more deeply into the other failing links...looks like some strange truncation of the url but not sure how this is happening.

fredguth:
@gr2m Worked fine as an app-specific plugin.  But couldn't follow the explanation on how to do the same with a npm module (3rd party plugin).

gr2m:
@fredguth 

> But couldn't follow the explanation on how to do the same with a npm module (3rd party plugin).

Can you tell what went wrong?

janl:
@gr2m works nearly, two snags:

1. https://github.com/hoodiehq/hoodie-plugin-hello-world#usage uses a `helloWorld` key in `package.json` `hoodie.app` which seems odd, should that be `hello-world`?

2. regardless of 1:
> you should see "Hello, world!" at http://localhost:8080/hoodie/\<your app name\>/api)

All variations of this throw a 404: 
- http://localhost:8080/hoodie/testapp/api (as per quickstart instructions and `package.json`
- http://localhost:8080/hoodie/hello-world/api (I would imagine the plugin name is a path prefix, so plugins don’t step over themselves?
- http://localhost:8080/hoodie/helloWorld/api (same as before, but taking 1. into consideration).

All of these show `{"statusCode":404,"error":"Not Found"}`.


* * *

Aside, I tried adding `blipp` to see the actual registered routes, but got intro trouble using the `hapiPlugins` array: https://github.com/hoodiehq/hoodie/pull/711/files#diff-c9bb31bbcc08d333b150cdd7e696ad9eR20 because when the actual `register()` is called, it gets passed the global `options` object which `blipp` doesn’t like:

```
/Users/jan/Work/tmp/testapp/node_modules/hoek/lib/index.js:731
    throw new Error(msgs.join(' ') || 'Unknown error');
    ^

Error: {
  "store" [1]: {
    "hooks": {}
  },
  "client" [2]: {},
  "admin" [3]: {},
  "account" [4]: {
    "admins": {},
    "secret": "b304c2256886b471c9d42f6148302b4d",
    "usersDb": "_users"
  },
  "db" [5]: {},
  "inMemory" [7]: false,
  "app" [8]: {
    "hello-world": {
      "greeting": "Bonjour",
      "name": "le monde"
    }
  },
  "plugins" [9]: [
    "@hoodie/plugin-hello-world"
  ],
  "paths" [10]: {
    "data": ".hoodie",
    "public": "/Users/jan/Work/tmp/testapp/node_modules/hoodie/public"
  },
  "loglevel" [11]: "warn",
  "name" [12]: "testapp"
}

[1] "name" is not allowed
[2] "loglevel" is not allowed
[3] "paths" is not allowed
[4] "plugins" is not allowed
[5] "app" is not allowed
[6] "inMemory" is not allowed
[7] "PouchDB" is not allowed
[8] "db" is not allowed
[9] "account" is not allowed
[10] "admin" is not allowed
[11] "client" is not allowed
[12] "store" is not allowed
    at Object.exports.assert (/Users/jan/Work/tmp/testapp/node_modules/hoek/lib/index.js:731:11)
    at Object.exports.register (/Users/jan/Work/tmp/testapp/node_modules/blipp/lib/index.js:20:10)
    at Object.target [as register] (/Users/jan/Work/tmp/testapp/node_modules/hapi/node_modules/joi/lib/types/object/index.js:77:34)
    at each (/Users/jan/Work/tmp/testapp/node_modules/hapi/lib/plugin.js:310:14)
    at iterate (/Users/jan/Work/tmp/testapp/node_modules/hapi/node_modules/items/lib/index.js:36:13)
    at done (/Users/jan/Work/tmp/testapp/node_modules/hapi/node_modules/items/lib/index.js:28:25)
    at each (/Users/jan/Work/tmp/testapp/node_modules/hapi/lib/plugin.js:261:28)
    at iterate (/Users/jan/Work/tmp/testapp/node_modules/hapi/node_modules/items/lib/index.js:36:13)
    at Object.exports.serial (/Users/jan/Work/tmp/testapp/node_modules/hapi/node_modules/items/lib/index.js:39:9)
    at internals.Plugin.register (/Users/jan/Work/tmp/testapp/node_modules/hapi/lib/plugin.js:313:11)
```

Of course I can register this manually and/or further up, but I’d say we should re-think that section of the code, if it only works with some hapi plugins.

gr2m:
I’m not sure how to make sth like `blipp` work to be honest, I’m not sure if it’s enough to just pass in options from say `options.app.<plugin name>`. Maybe one way we could make this work is by having the global options be passed to the `Hapi.Server({app: globalOptions})`, so plugins could still access them via `server.settings.app.*`?

janl:
I feel like treating hapi plugins and hoodie plugins the same leads to more trouble than is worth. E.g. if we want to load inert, and maybe at least for debugging blipp, maybe we should load them in ./server/index.js and not here in the hoodie-plugin loader.

gr2m:
@janl did you add `@hoodie/plugin-hello-world` to `hoodie.plugins[]` in your package.json file? Did you `npm link` it? If you did the latter, I just run into this myself and will pushed a fix via eb77206

gr2m:
nevermind, I removed the workaround. The right way to work with linked packages like this is to start hoodie with

```
NODE_PATH='/path/my/myhoodieapp' npm start
```

Or simpler: 

```
NODE_PATH=`pwd`/node_modules npm start
```

I’m still not sure about whether to pass only `hoodie.options['<plugin-name>']` to plugins or the entire options. The benefit if we would only pass in `hoodie.options['<plugin-name>']` then one could directly export hapi plugins at `./hoodie/server`, e.g.

```js
module.exports = require('blipp') // can be an array of plugins, too
```

I would not allow to add hapi plugins directly to the hoodie plugins array. Right now we enforce the folder structure by doing `reqire('<plugin-name>/hoodie/server')` so if we would add `blipp` to the array of plugin names it would try to `reqire('blipp/hoodie/server')` and it would fail. I’d say that’s okay. The alternative would be not to enforce the hoodie plugin / app folder structure any longer and I’d like to try to avoid that.

Something I forgot about is passing options to the client plugin code. Right now there is now way to do that. What I would suggest is to extend the `hoodie.plugin` method to pass a 2nd argument which is the options passed to the `@hoodie/client` constructor. That would allow us to do this

```js
// server/client/index.js
module.exports = function (hoodie, options) {
  hoodie.myMethod = function () {
    // do something with options here
  }
}
```

**Update:** I’ve started a PR which implements that https://github.com/hoodiehq/hoodie-client/pull/149. Together with https://github.com/hoodiehq/hoodie/pull/764 this implements passing client options

`options` would be what ever we pass as `options.client` to the `hoodie` hapi plugin (which is [currently not yet happening](https://github.com/hoodiehq/hoodie/blob/79dde6b8c4151d803aea0a1e58463d3ad6c06663/server/plugins/client/bundle.js#L85-L92) – I started a PR to fix that: https://github.com/hoodiehq/hoodie/pull/764). 

I guess it makes sense to follow the same convention for client option as we do for server option. Say I’d like to set a default greeting for my app, I’d do this in package.json

```json
{
  "name": "myhoodie",
  ...
  "hoodie": {
    "plugins": [
      "@hoodie/plugin-hello-world"
    ],
    "client": {
      "app": {
        "hello-world": {
          "greeting": "Bonjour"
        }
      }
    }
  }
}
```

An alternative would be to do `hoodie.app.<plugin name>.client` instead of `hoodie.client.app.<plugin name>`

```json
{
  "name": "myhoodie",
  ...
  "hoodie": {
    "plugins": [
      "@hoodie/plugin-hello-world"
    ],
    "app": {
      "hello-world": {
        "client": {
          "greeting": "Bonjour"
        }
      }
    }
  }
}
```

janl:
> @janl did you add @hoodie/plugin-hello-world to hoodie.plugins[] in your package.json file? Did you npm link it?

I added it to `package.json` as documented, and I did not `npm link` it

janl:
> I would not allow to add hapi plugins directly to the hoodie plugins array. 

I agree, but that’s not what I meant. I was considering the scenario of adding a hapi plugin to Hoodie core.

Seeing https://github.com/hoodiehq/hoodie/pull/711/files#diff-c9bb31bbcc08d333b150cdd7e696ad9eL24 I’d assume I can just add one there (which is what I tried), but that fails, because hapi plugins validate the options they are passed and the register code just throws in the whole Hoodie options object, which then makes the e.g. blipp registration fail.

`inert` just happens to work here.

gr2m:
> I added it to package.json as documented, and I did not npm link it

Then I don’t know what went wrong. Can you share your example hoodie app code, maybe create a repo for it so I can have a look?

> the register code just throws in the whole Hoodie options object

That is no longer the case via https://github.com/hoodiehq/hoodie/pull/711/commits/3c160a9dd586bdf2fec1a6515c20f655c62dbc52

janl:
> That is no longer the case via 3c160a9

@gr2m sounds good, thanks!

What about my earlier observations (copied from https://github.com/hoodiehq/hoodie/pull/711#issuecomment-305977797)

1. https://github.com/hoodiehq/hoodie-plugin-hello-world#usage uses a `helloWorld` key in `package.json` `hoodie.app` which seems odd, should that be `hello-world`?

2. regardless of 1:
> you should see "Hello, world!" at http://localhost:8080/hoodie/\<your app name\>/api)

All variations of this throw a 404: 
- http://localhost:8080/hoodie/testapp/api (as per quickstart instructions and `package.json`
- http://localhost:8080/hoodie/hello-world/api (I would imagine the plugin name is a path prefix, so plugins don’t step over themselves?
- http://localhost:8080/hoodie/helloWorld/api (same as before, but taking 1. into consideration).

All of these show `{"statusCode":404,"error":"Not Found"}`.

gr2m:
> All variations of this throw a 404:

have you seen https://github.com/hoodiehq/hoodie/pull/711#issuecomment-306033050

Can you push your test app somewhere? It all works for me

> uses a helloWorld key in package.json hoodie.app which seems odd, should that be hello-world?

the keys on `hoodie.app` don’t really matter if we just pass the whole options object, and `options.app.helloWorld` is easier to access than `options.app['hello-world']`

janl:
> have you seen #711 (comment)

yes: https://github.com/hoodiehq/hoodie/pull/711#issuecomment-306054759

gr2m:
let’s pair on this because I’m not able to reproduce the issue you are describing

gr2m:
can you show me the `package.json` file of your app and the contents of `hoodie/server.js` or `hoodie/server/index.js` depending on what you have?

gr2m:
regarding https://github.com/hoodiehq/hoodie/pull/711#issuecomment-305977797

> All variations of this throw a 404: 
>
> - http://localhost:8080/hoodie/testapp/api (as per quickstart instructions and `package.json`
> - http://localhost:8080/hoodie/hello-world/api (I would imagine the plugin name is a path prefix, so plugins don’t step over themselves?
> - http://localhost:8080/hoodie/helloWorld/api (same as before, but taking 1. into consideration).

The reason why `http://localhost:8080/hoodie/testapp/api` returned a 404 is that there was no `hoodie/server.js` file in the app. `http://localhost:8080/hoodie/hello-world/api` didn’t work because it should have been without the `/api`. I think when Jan tested it there was a typo in the instructions that has meanwhile been fixed.

So this all seems to work and is ready to be merged, I think. Needs a review accepting it from one of the other maintainers.

janl (approved):
+1

gr2m:
wheeeee I’m gonna merge this but would not yet publicly announce it. As we people can now technically start making their own plugins we should put some thought into that and give guidance

PR #745, opened 2017-05-25
[DO NOT MERGE] wip: de-es2015-ify #711

janl:
PR against #711 

I started porting the ES2015 code to ES5, with one `Object.assign()` left with a note: https://github.com/hoodiehq/hoodie/pull/745/files#diff-c9bb31bbcc08d333b150cdd7e696ad9eR21

Also, I found that the code assumes plugins being defined in `package.json` and getting “map is not a function on config.plugins` when non are defined. I added hot fixes for those, just to see if I get the rest right, that should be cleaned up: https://github.com/hoodiehq/hoodie/pull/745/files#diff-c9bb31bbcc08d333b150cdd7e696ad9eR29 and https://github.com/hoodiehq/hoodie/pull/745/files#diff-9f4a3658710e8d0ddf5ec999ae25f2adR31

### DIFF
--- a/server/plugins/index.js
+++ b/server/plugins/index.js
@@ -17,9 +17,9 @@ function checkModule (module) {
 }
 
 function registerPlugins (server, config, callback) {
-  var options = {
-    config: config
-  }
+
+  const options = Object.assign({ config }, config.pluginOptions)
+
   var hapiPlugins = [
     require('inert')
   ]
--- a/server/plugins/index.js
+++ b/server/plugins/index.js
@@ -32,7 +32,8 @@ function registerPlugins (server, config, callback) {
   ]
     .concat(
   [
-    path.resolve('hoodie/server')
+    path.resolve('hoodie/server'),
+    ...(config.plugins.map(i => `${i}/hoodie/server`))
   ]
     .filter(checkModule)
     )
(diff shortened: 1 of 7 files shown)


server/plugins/public.js
```diff
@@ -4,13 +4,13 @@ module.exports.register.attributes = {
...
-const path = require('path')
-const requireResolve = require('./resolver')
+var path = require('path')
+var requireResolve = require('./resolver')
 var createReadStream = require('fs').createReadStream
 var pathJoin = path.join
 
 function register (server, options, next) {
-  const { paths, plugins } = options.config
+  var { paths, plugins } = options.config
```
janl: Todo: no deconstructing

janl: Fixed

server/plugins/index.js
```diff
@@ -18,12 +18,16 @@ function checkModule (module) {
...
+  var options = Object.assign({ config: config }, config.pluginOptions)
 
   var hapiPlugins = [
     require('inert')
   ]
-
+  config.plugins = config.plugins || [] // doesn’t work, config.plugins is {} here
+  if (Object.keys(config.plugins).length === 0) {
+    config.plugins = [] // workaround for above issue
+  }
```
gr2m, line 30: I think that `config.plugins` should always be set, by setting it to `[]` in https://github.com/hoodiehq/hoodie/blob/master/cli/hoodie-defaults.js

gr2m:
we incorporated changes from this PR manually into #711

PR #747, opened 2017-05-25
remove obsolete options.db from tests

gr2m:


gr2m:
Merged this myself as it was minor without any logic changes

PR #748, opened 2017-05-28
fix local test coverage on Macs

gr2m:
Without the quotes, the glob is resolved by the OS, at least on Macs. Because of that, the `./test/unit/plugins-client-bundle-test.js` test was not run and the coverage fell when run locally, while it was still 100% on travis


gr2m:
as this is very minor, I’ll merge it myself when CI is green. But please ask me anything about it even when merged :)

PR #752, opened 2017-05-30
Issue 751 Fail gracefully when package.json cannot be found

### LINKED ISSUE #751: Hoodie install breaks on my (admittedly unusual) setup

Steps to reproduce:
```
mkdir proxy
mkdir proxy/node_modules
mkdir app
cd app
npm init -y
npm i -S hoodie
```

parallaxeffect:
Catch the error thrown by require when setup.js cannot find the parent package.json. This can happen if there are symbolic links involved in the main project's setup.

closes #751

### DIFF
--- a/bin/setup.js
+++ b/bin/setup.js
@@ -4,9 +4,10 @@ var path = require('path')
 var fs = require('fs')
 var log = require('npmlog')
 
-var installIntoApp = process.env.PWD.indexOf('node_modules') !== -1
+var isInstallIntoApp = /node_modules/.test(process.cwd())
+
 // This block only executes if Hoodie is installed with as a dependency
-if (installIntoApp) {
+if (isInstallIntoApp) {
   var pathToAppRoot = path.resolve('..', '..')
   var packageJson = require(path.join(pathToAppRoot, 'package.json'))
 


gr2m:
Thanks @parallaxeffect! Sorry that nobody got to review your PR yet, we got it on our list :)

parallaxeffect:
Sure, no problem.
It's possible you might want to add a more informative error message. This fix prevented the install from failing, but it did take me a while to figure out how to get hoodie running from there. You need to either manually add hoodie to your scripts, or run it out of the .bin/ folder.

gr2m:
Hey @parallaxeffect, I think the solution you suggested in https://github.com/hoodiehq/hoodie/issues/751#issue-232087182 sounds more elegant. I don’t think it will break anything. I pushed a branch and started a PR (https://github.com/hoodiehq/hoodie/pull/761) to see if it goes trough the quickstart test. If that works for you, too, I’d suggest we go with that?

bin/setup.js
```diff
@@ -12,31 +12,40 @@ var saveRequested = process.env.npm_config_save
...
+          return
+        }
+        log.info('setup', 'Start your Hoodie app with "npm start"')
+      })
+  } catch (error) {
+    if (error.code !== 'MODULE_NOT_FOUND') {
+      throw error
+    }
+    log.error('setup', 'Could not find package.json at ' + path.join(pathToAppRoot, 'package.json'))
+    log.error(error)
```
gr2m: Maybe a warning would be enough? 

```js
log.warn('setup', 'Could not find package.json at ' + path.join(pathToAppRoot, 'package.json')
log.warn('setup', 'You must manually set start script in your app’s package.json to "hoodie" in order for "npm start" to work')
```

parallaxeffect: Yeah, a warning should be sufficient. Not really familiar with the logging system.

gr2m:
wait this was not meant to be closed, I pushed a bad commit to your fork, sorry! I’ll made a follow up PR: #766

PR #753, opened 2017-05-31
doc: allow 404 in link checker

fozy81:
avoid false positives issue in link checker

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -15,8 +15,9 @@ before_script:
   # remove following files to exclude from awesome_bot link checks because '127.0.0.'... links cause failure even when whitelisted
   - rm docs/guides/quickstart.rst.md docs/guides/deployment.rst.md
   - gem install awesome_bot
-  # run dead link checks but silence 404 failures (due to false postives), redirect, dupes and excluding white-listed links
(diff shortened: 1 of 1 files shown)


.travis.yml
```diff
@@ -15,8 +15,8 @@ before_script:
   # remove following files to exclude from awesome_bot link checks because '127.0.0.'... links cause failure even when whitelisted
   - rm docs/guides/quickstart.rst.md docs/guides/deployment.rst.md
   - gem install awesome_bot
-  # run dead link checks excluding white-listed links
-  - awesome_bot --allow-redirect --allow-dupe docs/**/*.md README.md --white-list http://127.0.0.1:8080,http://myhoodieapp.com,https://pouchdb.com/adapters,https://hoodie-app-tracker-randomxyz.now.sh/,https://my-tracker-app.now.sh,https://hapijs.com/tutorials/,https://opencollective.com
+  # run dead link checks but silence 404 failures (due to false postives), redirect, dupes and excluding white-listed links
+  - awesome_bot --allow 404 --allow-redirect --allow-dupe docs/**/*.md README.md --white-list http://127.0.0.1:8080,http://myhoodieapp.com,https://pouchdb.com/adapters,https://hoodie-app-tracker-randomxyz.now.sh/,https://my-tracker-app.now.sh,https://hapijs.com/tutorials/,https://opencollective.com
```
gr2m: I don’t think it’s only 404s. For example jsbin tends to go down from time to time which results in a connection error, I don’t think it’s treated as 404 in that case. Any thought?

fozy81: Added commit to allow no response, timeout, bad gateway, service unavailable and gateway timeout. That should cover most common issues. But maybe this is getting a little complex?

gr2m (approved):

gr2m:
Thanks @fozy81

Issue #758, opened 2017-06-03
Get /hoodie/client.js fails with npm2


gr2m:
We bundle the hoodie `client.js` dynamically on first request to `GET /hoodie/client.js`, see https://github.com/hoodiehq/hoodie/blob/2bc4c6da06c3b5efb8f9005c534a439d08a5ae7b/server/plugins/client/bundle.js#L91

The `require('pouchdb-browser')` fails in npm2 because it’s a sub dependency and does not exist in `node_modules/pouchdb-browser` but `node_modules/hoodie/node_modules/pouchdb-browser` instead

I’m not sure if there is a clever way to work around that, so that we remain compatible with npm2, or if we give up on it. The work around is to install `pouchdb-browser` and `@hoodie/client` as dependencies of the app itself:

```bash
npm install --save pouchdb-browser @hoodie/client
```

PR #759, opened 2017-06-03
Quickstart test: fail when server responds with error

### LINKED ISSUE #749: CI: test pass even though bin/quick-start-test.sh fails in `before_script`
see https://travis-ci.org/hoodiehq/hoodie/builds/236919898#L599

For one, the error doesn’t make sense to me, `pouchdb-browser` was installed above so I’m not sure what is happening here. Maybe a problem with npm@2, I can’t test that right now

But more importantly, the tests should fail if the `bin/quick-start-test.sh` script fails

### Update

It is in fact a problem with npm@2: https://github.com/hoodiehq/hoodie/issues/758

gr2m:
closes #749

### DIFF
--- a/bin/quick-start-test.sh
+++ b/bin/quick-start-test.sh
@@ -6,14 +6,22 @@ trap "{ CODE=$?; popd; rm -rf $TEMP_ROOT; exit $CODE; }" EXIT
 
 ## docs/guides/quickstart.rst:41
 
+echo "$ npm init -y"
 npm init -y
 
 ## docs/guides/quickstart.rst:47
 
+echo "$ npm install $HOODIE_FOLDER --save"
 npm install $HOODIE_FOLDER --save
+
+## docs/guides/quickstart.rst:98
+
+echo "$ npm install --save pouchdb-browser @hoodie/client"
+npm install --save pouchdb-browser @hoodie/client
 
 ## docs/guides/quickstart.rst:72
 
+echo "$ npm start"
 npm start & HOODIE_PROCESS=$!
 
 ## Give the process a generous sixty seconds to start
--- a/bin/quick-start-test.sh
+++ b/bin/quick-start-test.sh
@@ -25,14 +25,16 @@ echo "$ npm start"
 npm start & HOODIE_PROCESS=$!
 
 ## Give the process a generous sixty seconds to start
-
 RETRIES=60
 while [ "$RETRIES" -gt 0 ] ; do
   sleep 1
   kill -0 $HOODIE_PROCESS 2&>1 > /dev/null || { echo "Hoodie exited prematurely"; wait $HOODIE_PROCESS ; exit $? ; }
-  curl -fI http://localhost:8080/hoodie/client.js > /dev/null && break
+  curl -fI http://localhost:8080 > /dev/null && break
   let RETRIES=RETRIES-1
 done
+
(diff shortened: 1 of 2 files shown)


gr2m:
@janl that worked: https://travis-ci.org/hoodiehq/hoodie/builds/239092845. But it still tries it 60 times instead of stop trying after the first fail.

Maybe there is no workaround for it because for the first few tries the server might not be up yet. I’d say this is good enough now. I’ll put in the workarounds from #758 so that the quickstart test passes again

gr2m:
This is ready for review, ping @hoodiehq/maintainers

janl:
`man curl` again:

```
       --retry <num>
              If a transient error is returned when curl tries to perform a transfer, it will
              retry this number of times before giving up. Setting the number to 0 makes curl
              do  no retries (which is the default). Transient error means either: a timeout,
              an FTP 4xx response code or an HTTP 5xx response code.

              When curl is about to retry a transfer, it will first wait one second and  then
              for all forthcoming retries it will double the waiting time until it reaches 10
              minutes which then will be the delay between the rest of the retries.  By using
              --retry-delay you disable this exponential backoff algorithm. See also --retry-
              max-time to limit the total time allowed for retries. (Added in 7.12.3)

              If this option is used several times, the last one will be used.

       --retry-delay <seconds>
              Make curl sleep this amount of time before  each  retry  when  a  transfer  has
              failed  with  a  transient error (it changes the default backoff time algorithm
              between retries). This option is only interesting if --retry is also used. Set-
              ting this delay to zero will make curl use the default backoff time.  (Added in
              7.12.3)

              If this option is used several times, the last one will be used.

       --retry-max-time <seconds>
              The retry timer is reset before the first transfer  attempt.  Retries  will  be
              done  as  usual  (see  --retry)  as long as the timer hasn't reached this given
              limit. Notice that if the timer hasn't reached the limit, the request  will  be
              made  and  while performing, it may take longer than this given time period. To
              limit a single request's maximum time, use -m, --max-time.  Set this option  to
              zero to not timeout retries. (Added in 7.12.3)

              If this option is used several times, the last one will be used.
```

gr2m:
`curl --retry 60 http://localhost:8080` fails right away without retrying. Am I doing something wrong?

```bash
$ curl --retry 60 http://localhost:8080
curl: (7) Failed to connect to localhost port 8080: Connection refused
```

janl:
> If a transient error is returned when curl tries to perform a transfer, it will retry this number of times before giving up.

& 

> Transient error means either: a timeout, an FTP 4xx response code or an HTTP 5xx response code.

Together this means these settings won’t apply when the host is not reachable, but *only* when you get a timeout, or HTTP 5xx response. But not for a connection refused, or 4xx error.

gr2m:
seems like curl can’t handle connection refused errors yet. It was brought up at https://github.com/curl/curl/issues/1036, but was not yet implemented as it seems. Probably that’s why @benwhite-deltas implemented the manual loop :)

gr2m:
@janl I added a change: https://github.com/hoodiehq/hoodie/commit/4ab9daf4a8ad63ed488a7a7afac45d2f883465a9

That way we have the loop that waits until the server starts. Only then we check if the client bundling works. What do you think?

gr2m:
ping @hoodiehq/maintainers this is ready for review

janl (approved):
+1

PR #761, opened 2017-06-03
support hoodie install with linked node_modules folder

### LINKED ISSUE #751: Hoodie install breaks on my (admittedly unusual) setup

gr2m:
closes #751. 

To test it

```
mkdir proxy
mkdir proxy/node_modules
mkdir app
cd app
npm init -y
npm i -S hoodiehq/hoodie#751/setup-with-linked-node-modules-folder
```

I just tried it with my computer (node v6.9.5, npm 3.10.10) and it failed with the same error as descried in #751. Did I miss something @parallaxeffect?

### DIFF
--- a/bin/setup.js
+++ b/bin/setup.js
@@ -4,10 +4,11 @@ var path = require('path')
 var fs = require('fs')
 var log = require('npmlog')
 
-var installIntoApp = process.env.PWD.indexOf('node_modules') !== -1
+var workingDirectory = process.env.PWD
+var installIntoApp = /node_modules/.test(workingDirectory)
 // This block only executes if Hoodie is installed with as a dependency
 if (installIntoApp) {
-  var pathToAppRoot = path.resolve('..', '..')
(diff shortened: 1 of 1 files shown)


parallaxeffect:
You're not missing anything. I get the same result. npm must be doing something strange that I don't understand. Note that the new version works if run manually with `node ./bin/setup.js` or `npm run postinstall` as long as they are run from `app/node_modules/hoodie` and not `proxy/node_modules/hoodie`, whereas the previous version failed on both.
I'm not sure why npm acts differently during the install process.
There is some information on what looks like a related situation on stackoverflow: https://stackoverflow.com/questions/22425395/node-js-require-does-not-look-in-parent-directory-when-using-npm-link

gr2m:
shall we close this PR as it doesn’t seem to resolve the problem of #751?

parallaxeffect:
Yeah. Feel free to.

PR #764, opened 2017-06-04
pass `options.client` to `Hoodie` client constructor

### LINKED ISSUE #762: merge `options.client` of hapi plugin into bundled `new Hoodie(options)`
see https://github.com/hoodiehq/hoodie/blob/79dde6b8c4151d803aea0a1e58463d3ad6c06663/server/plugins/client/bundle.js#L85-L92

what we want is stringify `options.client` and then set `option.url` from the hapi plugin’s `options.url`

gr2m:
closes #762


janl:
+1

PR #765, opened 2017-06-04
windows compatibility for postinstall setup

### LINKED ISSUE #763: process.env.PWD is not set on Windows
follow up for #760. See https://stackoverflow.com/questions/44355912/hoodie-postinstall-failed-process-env-pwd-is-undefined ping @janl

gr2m:
closes #763

### DIFF
--- a/bin/setup.js
+++ b/bin/setup.js
@@ -4,9 +4,10 @@ var path = require('path')
 var fs = require('fs')
 var log = require('npmlog')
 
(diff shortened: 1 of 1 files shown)


gr2m:
merging because it’s currently breaking install of Hoodie on Windows. I test it with npm@5 and `npm i hoodiehq/hoodie#763/windows-postinstall-setup-fix` and it worked

janl:
retroactive +1, thanks for moving so quickly with this!

PR #766, opened 2017-06-05
setup: fail gracefully if package.json cannot be found

### LINKED ISSUE #751: Hoodie install breaks on my (admittedly unusual) setup

gr2m:
closes #751, follow up for #752

### DIFF
--- a/bin/setup.js
+++ b/bin/setup.js
@@ -9,7 +9,19 @@ var isInstallIntoApp = /node_modules/.test(process.cwd())
 // This block only executes if Hoodie is installed with as a dependency
 if (isInstallIntoApp) {
   var pathToAppRoot = path.resolve('..', '..')
-  var packageJson = require(path.join(pathToAppRoot, 'package.json'))
+
+  // log warning if package.json cannot be found (hoodiehq/hoodie#751)
+  try {
+    var packageJson = require(path.join(pathToAppRoot, 'package.json'))
+  } catch (error) {
+    if (error.code !== 'MODULE_NOT_FOUND') {
+      throw error
+    }
+    log.warn('setup', 'Could not find package.json at ' + path.join(pathToAppRoot, 'package.json'))
+    log.warn('setup', 'You must manually set the start script in your app’s package.json to "hoodie" in order for "npm start" to work')
+
+    process.exit(0)
+  }
 
   packageJson.scripts = packageJson.scripts || {}
 
--- a/bin/setup.js
+++ b/bin/setup.js
@@ -6,8 +6,12 @@ var log = require('npmlog')
 
 var isInstallIntoApp = /node_modules/.test(process.cwd())
 
-// This block only executes if Hoodie is installed with as a dependency
+// Run setup only if Hoodie is installed explicitly (npm install hoodie)
 if (isInstallIntoApp) {
+  setup()
+}
+
+function setup () {
   var pathToAppRoot = path.resolve('..', '..')
 
   // log warning if package.json cannot be found (hoodiehq/hoodie#751)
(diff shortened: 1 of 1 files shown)


gr2m:
this is ready for review ping @hoodiehq/maintainers

bin/setup.js
```diff
@@ -9,7 +9,19 @@ var isInstallIntoApp = /node_modules/.test(process.cwd())
...
+  try {
+    var packageJson = require(path.join(pathToAppRoot, 'package.json'))
+  } catch (error) {
+    if (error.code !== 'MODULE_NOT_FOUND') {
+      throw error
+    }
+    log.warn('setup', 'Could not find package.json at ' + path.join(pathToAppRoot, 'package.json'))
+    log.warn('setup', 'You must manually set the start script in your app’s package.json to "hoodie" in order for "npm start" to work')
+
+    process.exit(0)
```
janl: I seem to be remembering a somewhat recent major version change in Node.js that `process.exit` might cause `console.log` messages to be swallowed because of some async operation. I think this note in https://nodejs.org/api/process.html#process_a_note_on_process_i_o refers to what I mean:

> Synchronous writes avoid problems such as output written with console.log() or console.error() being unexpectedly interleaved, or not written at all if process.exit() is called before an asynchronous write completes. See [process.exit()](https://nodejs.org/api/process.html#process_process_exit_code) for more information.

and from https://nodejs.org/api/process.html#process_process_exit_code

> It is important to note that calling process.exit() will force the process to exit as quickly as possible even if there are still asynchronous operations pending that have not yet completed fully, including I/O operations to process.stdout and process.stderr

Now the question here is if `log.warn()` is sync or async.

If they are, I think the solution is to register with the `beforeExit` event on `process` and do the desired output there: https://nodejs.org/api/process.html#process_event_beforeexit

* * *

And a matter of style, if `exit`ing because of an error condition, a return code other than `0` (zero) should be used. `1` would be an option.

gr2m: it’s not an error, the install already happened. It’s only the automated setup which fails and we log instructions on how to manually do it instead, so I thought warning fit better?

`log.warn()` is synchronous, we don’t write any file, I think this should be fine

janl: > we don’t write any file

stdout is a file :)

> log.warn() is synchronous

What does it use under the hood?

gr2m: looks like they use `process.stderr.write` (https://github.com/npm/npmlog/blob/master/log.js#L12, https://github.com/npm/npmlog/blob/master/log.js#L266)

gr2m: I didn’t know that `console.log`s or `process.(stderr|stdout).write`s could potentially not finish if we call process.exit(), I think I do that quite a bit. I’d love to learn what the proper way to handle this is. I’m not sure if a `beforeexit` event handler will help because the documentation says 

> The 'beforeExit' event is not emitted for conditions causing explicit termination, such as calling process.exit()

janl: I misread the `beforeExit`, sorry!

https://www.nczonline.net/blog/2014/02/04/maintainable-node-js-javascript-avoid-process-exit/ explains the issue.

https://github.com/nodejs/node-v0.x-archive/issues/8329#issuecomment-54778937 Suggests wrapping the `process.exit()` call in a `process.nextTick()`

gr2m: okay I read trough https://github.com/nodejs/node/issues/6456 this is really complicated, now I understand npm’s frustration around Node@6’s release.

I pushed another commit to workaround using `process.exit()` altogether: 
https://github.com/hoodiehq/hoodie/pull/766/commits/182a818ff93a10ab14a2590ba6bf0284dd899239

gr2m:
fixed via https://github.com/hoodiehq/hoodie/commit/79dde6b8c4151d803aea0a1e58463d3ad6c06663

PR #773, opened 2017-06-09
hapi security update

gr2m:
see https://nodesecurity.io/advisories/hapi_denial-of-service-via-malformed-accept-encoding-header


gr2m:
going to merge because it’s a security update and has no change anyway for most people

PR #780, opened 2017-07-02
package.json updated

### LINKED ISSUE #779: node_modules/hoodie folder is 23,8mb
image

nearly all of it are unneeded things like the pandoc binary and docs. We should remove it by adding the ["files" setting](https://docs.npmjs.com/files/package.json#files) to `hoodie`’s package.json file and only list what’s needed to be published

sohini-roy:
closes #779

### DIFF
--- a/package.json
+++ b/package.json
@@ -11,6 +11,9 @@
     "url": "https://github.com/hoodiehq/hoodie/issues"
   },
   "dependencies": {
+    "@hoodie/admin": "^1.0.1",
+    "@hoodie/client": "^10.0.0",
+    "@hoodie/server": "^23.0.0",
     "async": "^2.0.0",
     "browserify": "^14.0.0",
     "good": "^7.0.2",
--- a/package.json
+++ b/package.json
@@ -11,9 +11,6 @@
     "url": "https://github.com/hoodiehq/hoodie/issues"
   },
   "dependencies": {
-    "@hoodie/admin": "^1.0.1",
-    "@hoodie/client": "^10.0.0",
-    "@hoodie/server": "^23.0.0",
     "async": "^2.0.0",
     "browserify": "^14.0.0",
     "good": "^7.0.2",
@@ -53,6 +50,7 @@
   "engines": {
     "node": ">=4"
   },
+  "files":["@hoodie/admin", "@hoodie/client", "@hoodie/server"],
   "homepage": "https://github.com/hoodiehq/hoodie#readme",
   "keywords": [
     "nobackend",
--- a/package.json
+++ b/package.json
@@ -53,6 +53,9 @@
   "engines": {
     "node": ">=4"
   },
+  "files": [
+    "bin", "cli", "public", "server"
+  ],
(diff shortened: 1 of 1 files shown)


package.json
```diff
@@ -53,6 +50,7 @@
   "engines": {
     "node": ">=4"
   },
+  "files":["@hoodie/admin", "@hoodie/client", "@hoodie/server"],
```
gr2m: the `"files"` key in the `package.json` is not for names of dependencies, but for file names and paths that are part of the repository, please read the description at that’s not how you use `files`. See the documentation at https://docs.npmjs.com/files/package.json#files

gr2m (changes requested):

package.json
gr2m: e.g. see https://github.com/hoodiehq/hoodie-admin/blob/master/package.json#L14-L16. This means that when the package get published to the npm registry, only the contents of the `dist/` folder get uploaded, everything else is ignored. Does that  make sense

sohini-roy: i did follow that. i am sure i misunderstood but i am still confused.

gr2m: when you install `hoodie` like it’s described in the [quickstart guide](http://docs.hood.ie/en/latest/guides/quickstart.html) and then open `node_modules/hoodie`, you will see a ton of files in there, including the pandoc file which is 20mb of size. Hoodie doesn’t need that file to work, nor does it need e.g. the docs folder. The only file it needs are the JavaScript files that get executed when you run `npm start` in your app. These are `bin/`, `cli/` and `server/`. And then it also needs the `public/` folder as it has the default UI that will be shown when your app has no `own` public folder. So the names of these 4 folders are the one that we need to to set `"files"` to in our `package.json`. That way only these folders will be uploaded to the npm registry, all other files will be ignored. Does that make sense?

sohini-roy: yes i get it

sohini-roy:
please have a look and review

package.json
```diff
@@ -11,9 +11,6 @@
     "url": "https://github.com/hoodiehq/hoodie/issues"
   },
   "dependencies": {
-    "@hoodie/admin": "^1.0.1",
-    "@hoodie/client": "^10.0.0",
-    "@hoodie/server": "^23.0.0",
```
gr2m: you need to put these back in, that’s why the tests are currently failing, see https://travis-ci.org/hoodiehq/hoodie/builds/249421274#L564

package.json
```diff
@@ -53,6 +50,9 @@
   "engines": {
     "node": ">=4"
   },
+  "files": [
+    "bin", "cli", "public", "server"
+  ],
```
gr2m, line 58: that looks good 👍

gr2m (changes requested):

package.json
sohini-roy: thanks for helping :)

gr2m (approved):
great, thanks :)

gr2m:
and released :) https://github.com/hoodiehq/hoodie/releases/tag/v28.2.1

PR #799, opened 2017-09-04
Wrote stuff about the name property

### LINKED ISSUE #793: Configuration Docs: "name" argument is not documented
By default, the app name is the package.json "name" value. But it can be overwritten in `"hoodie": { "name": "my-custom-name"}`

maraisr:
Resolves #793


docs/guides/configuration.rst
```diff
@@ -33,6 +33,8 @@ Here is a list of all available options
 +-----------------+--------------------------+--------------------------+--------------------------+---------------------------------------------------------------------------------------------------------------------+
 | adminPassword   | -                        | ``--adminPassword``      | ``hoodie_adminPassword`` | Password to login to Admin Dashboard. Login is not possible unless set                                              |
 +-----------------+--------------------------+--------------------------+--------------------------+---------------------------------------------------------------------------------------------------------------------+
+| name            | hoodie                   | ``--name``               | ``hoodie_name``          | Name your application.                                                                                              |
++-----------------+--------------------------+--------------------------+--------------------------+---------------------------------------------------------------------------------------------------------------------+
```
gr2m: the default value is not `hoodie`, but the value of the `"name"` property in the `package.json` file

docs/guides/configuration.rst
```diff
@@ -54,7 +56,8 @@ default values
         "inMemory": false,
         "loglevel": "warn",
         "url": "",
-        "adminPassword": ""
+        "adminPassword": "",
+        "name": "hoodie"
```
gr2m: maybe use `"my-hoodie-app"` as an example value? `"hoodie"` might be confusing

gr2m (changes requested):
Thank you Marais! I’ve two comments, could you please look into it?

docs/guides/configuration.rst
maraisr: :boom: thanks man - mod going up shortly!

docs/guides/configuration.rst
maraisr: Definitely agree :sunglasses:! Thanks man.

gr2m (approved):
Looks great, thanks for the changes 👍

PR #804, opened 2017-09-19
Fix formatting issue in files-and-folders.rst

garbados:
Due to some extra spaces, a list appears malformed. This PR removes those extra spaces to preserve correct formatting.


gr2m (approved):
Thanks again :)

gr2m:
Are you on twitter by chance? We like giving shout outs to new contributors but no worries if prefer us not to :)

garbados:
I'm not on twitter but i'm on mastodon! https://toot.cat/@garbados

PR #808, opened 2017-09-27
Add documentation for dbUrl* options.

garbados:
I encountered this error while attempting to use `--adminPassword`:

```
Error: Authentication credentials (username AND password) are missing from dbUrl. Provide them either by --dbUrl directly or by setting --dbUrlUsername and --dbUrlPassword arguments.
```

I couldn't find documentation elsewhere on `dbUrlUsername` and `dbUrlPassword`, so I added some.


docs/guides/configuration.rst
```diff
@@ -19,6 +19,10 @@ Here is a list of all available options
...
 +-----------------+----------------------------------+--------------------------+--------------------------+---------------------------------------------------------------------------------------------------------------------+
+| dbUrlUsername   | –                                | ``dbUrlUsername``        | ``hoodie_dbUrlUsername`` | If ``dbUrl`` is set, you can use ``dbUrlUsername`` to set the username to use when making requests to CouchDB       |
++-----------------+----------------------------------+--------------------------+--------------------------+---------------------------------------------------------------------------------------------------------------------+
+| dbUrlPassword   | –                                | ``dbUrlPassword``        | ``hoodie_dbUrlPassword`` | If ``dbUrl`` is set, you can use ``dbUrlPassword`` to set the password to use when making requests to CouchDB       |
++-----------------+----------------------------------+--------------------------+--------------------------+---------------------------------------------------------------------------------------------------------------------+
```
gr2m: Great catch! Can you adopt the description for `dbUrl`? So instead of "URL has to contain credentials" we could say sth like "URL has to contain credentials, or `dbUrlUsername` and `dbUrlPassword` have to be set separately"

garbados: Makes sense. I'll modify the language and push the change.

gr2m (approved):
thanks for the great PR :)

PR #810, opened 2017-09-30
Update style.css

keangkai:
add comment for easy to read

### DIFF
--- a/public/style.css
+++ b/public/style.css
@@ -27,6 +27,7 @@ h1 {
   text-align:center;
 }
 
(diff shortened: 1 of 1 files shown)


public/style.css
```diff
@@ -27,6 +27,7 @@ h1 {
   text-align:center;
 }
 
+/*Uppercase color orange*/
```
gr2m, line 30: what do you mean with uppercase? Maybe add a comment /* orange */ after `color:#e94e1b;` if that’s what makes it easier to read?

PR #811, opened 2017-10-02
Update 404.html

keangkai:
add some comment for easy to understand

### DIFF
--- a/public/404.html
+++ b/public/404.html
@@ -1,12 +1,14 @@
 <!DOCTYPE html>
 <html lang="en">
   <head>
+    /*define char same all page*/
     <meta charset="utf-8">
     <title>Hoodie</title>
-
+    /*internal css*/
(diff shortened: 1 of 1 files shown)


public/404.html
```diff
@@ -1,12 +1,14 @@
...
   <head>
+    /*define char same all page*/
     <meta charset="utf-8">
     <title>Hoodie</title>
-
+    /*internal css*/
     <link rel="stylesheet" type="text/css" href="style.css">
   </head>
   <body>
+    /*bigest char*/
```
gr2m: note that `/* */` are JavaScript comments :) For HTML comments you do `<!-- -->` :) 

And in this case, I think the comments are not needed, the HTML code is very simple already, don’t you think?

gr2m (changes requested):

public/404.html
keangkai: Sorry.I'm forget

PR #813, opened 2017-10-04
Update options.js

keangkai:
change comment for easy to read


gr2m (changes requested):
@keangkai I had questions about all your past pull requests. Please, can you stop sending pull requests without creating an issue about what you want to change first? 

If you send another one without reacting to my questions, I’ll have to block you from further participation on Hoodie, I’m sorry

PR #814, opened 2017-10-17
mytry

sthiagu:
Closes #801


gr2m:
pleas stop spamming for Hacktoberfest, the PRs are not counting and your spam account will get blocked.

PR #828, opened 2017-12-13
add meta require for search engine

keangkai:
Tell a word about this web page and tell the details of the webpage.

### DIFF
--- a/public/index.html
+++ b/public/index.html
@@ -2,6 +2,8 @@
 <html lang="en">
   <head>
     <meta charset="utf-8">
+    <meta name="keywords" content="">
+    <meta name="description" content="">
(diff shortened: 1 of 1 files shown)


gr2m (changes requested):
These keywords are not required by search engines, but they can help. Adding them without setting anything in `content=""` isn’t very helpful though? If anything, I would suggest we add a meta tag that prevents this page from being indexed by search engines, as it’s really just a placeholder. What do you think?

keangkai:
It’s good.first I will edit it and sorry

gr2m:
no worries at all, you can update this pull request or close it and create a new one. Let me know if you have any questions

PR #830, opened 2017-12-14
Update 404.html

Baimaikp:
If you are browsing this page with a phone or a tablet, you can click on the two links below to see the difference.

for easy to view in your phone


gr2m (approved):
👍 thank you

PR #852, opened 2018-04-17
I'm assuming there are only two o's in Hoodie

thomasvs:


gr2m (approved):
you assumed correctly 👌

PR #860, opened 2018-08-25
docs: update hoodie tracker app url

fozy81:
The web address for hoodie tracker has been change, this will update to correct address.


gr2m (approved):
👍

PR #862, opened 2018-11-12
update PouchDB’s requirements docs

DiogenesPolanco:
I would like to contribute with the project and this is my first contribution with the documentation because I am still studying the way they work with the project.


gr2m (approved):
Nice one, thanks :) Welcome to Hoodie!

PR #866, opened 2019-03-26
Update plugins.rst

drodsou:
As a beginner I spent some time and frustration until I realized where \<app name\> was coming from in the server plugins section.


gr2m (approved):
Sorry for the frustration :( thanks for improving the docs as a result 👍

PR #867, opened 2019-03-26
Fix client plugins crashing hoodie server in Windows

drodsou:
In Windows, adding a client plugin as described in `http://docs.hood.ie/en/latest/guides/plugins.html` will crash hoodie server with error:

> Error: Cannot find module 'C:UsersUserdevhoodie-serverhoodieclient' 

as backslashes in Windows paths are collapsed when building the string `hoodieBundleSource`. 

The fix replaces backslashes in the client plugins paths with forward slashes, which work both in Linux and Windows.


gr2m:
Thanks David!

PR #868, opened 2019-04-18
Call to eval-like DOM function

matbrgz:
DOM functions that act like 'eval' and execute strings as code are dangerous and impede program analysis and understanding. Consequently, they should not be used.


gr2m (approved):
fair enough :) thanks!

Issue #879, opened 2019-09-14
`options.account` and `options.store` not passed to hoodie-server


laurensiusadi:
Initially I was trying the https://github.com/hoodiehq/hoodie-app-tracker/ repo, and find it can't do `password reset` request. So I go ahead and post an issue there https://github.com/hoodiehq/hoodie-app-tracker/issues/187 ( a year ago)

Looking at the error thrown, it was from `hoodie-account-server`, so I though I'd move the issue there https://github.com/hoodiehq/hoodie-account-server/issues/302 (yesterday)

The main problem was `hoodie` doesn't seem to read options I put on `.hoodierc` file. But if I put `port` options there, `hoodie` port will change on run, so the file is read, no problem.

The problem then was `options.account` and `options.store` as in https://github.com/hoodiehq/hoodie-server wan't recognized or passed to the server register function.

Last night I found a breakthrough, I was looking at the wrong repo for the whole time for `rc` (was looking on account-server, server, client inside the node_modules under `@hoodie` folder), it should be on this main hoodie. The culprit is the options isn't passed to `Hoodie` constructor.

PR #880, opened 2019-09-14
fix: pass `options.account` and `options.store`

laurensiusadi:
Closing https://github.com/hoodiehq/hoodie/issues/879


laurensiusadi:
I messed up by committing the `package-lock.json`, but is this okay now? It will be squash merge right?

gr2m (approved):
Thanks for looking into this so thoroughly Laurensius! 

It's been so long that I looked into the code but I see no harm accepting the pull request if it resolves your problem

PR #882, opened 2019-10-05
changing docs

swagatikapanda123:


gr2m:
Hey thanks for trying to help, but I don't think your changes are an improvement over what we have today?

PR #883, opened 2019-10-12
Update glossary.rst

ateebahmad321:


docs/about/glossary.rst
```diff
@@ -30,7 +30,7 @@ Users
 Hoodie isn’t a CMS, but a backend for web apps, and as such, it is very
 much centered around users. All of the offline and sync features are
 specific to each individual user’s data, and each user’s data is
-encapsulated from that of all others by default. This allows Hoodie to
+encapsulated from of all others by default. This allows Hoodie to
```
gr2m, line 33: I think I would remove both "that" and "of"

```suggestion
encapsulated from all others by default. This allows Hoodie to
```

PR #886, opened 2019-10-21
updated npm depricated version to updated one

abdulhaqshah:
didn't updated the hapi because this need code changes and the latest version have an error that's why.


gr2m:
As we update the dependencies, we should also drop support for outdated Node versions, which would be a breaking version. I don't mind doing that, upgrading should be easy enough. 

Can you update 

https://github.com/hoodiehq/hoodie/blob/159f339b4237b0d8ae112aa0f6d9ade1059e7492/package.json#L52-L54

(Bump it to 8 for now)

And the `.travis.yml` file to test in Node 12 (which is LTS now)?

PR #892, opened 2019-11-07
updated dev dependencies and updated the way standard pretest is called

dhuang612:
completed the following:
```
nock updated from v9  =>  v11.7.0

proxyquire updated from v1.7.4 => v2.1.3

semantic-release updated from v12.2.2 => v14.0.2
if you try to bring this version up to v 15 there are new high risk vulnerabilities that get added related to man in the middle attacks

standard updated from v11.0.0 => 12.0.1
if you try to bring this closer to v 14 you run into hapi deprecation warnings related to the url.
changes to the code required.

Also updated the package.json call for npm pretest following the instruction from here:
https://standardjs.com/
"pretest": "standard --fix"
When you update the npm package standard it causes eslint errors to fail the code format.
By calling --fix when you run npm test it automatically resolves those errors for you.


tap updated from v11.0.0 => v12.1.1
If you try to bring it to a more recent version there are testing errors related to node v6

textlint updated from v10.0.0 => v11.5.0
```

successfully ran npm test which completed without errors.

Please let me know if anything else needs to be changed!


dhuang612:
Some additional thoughts:

Did you still want this completed?
```
As we update the dependencies, we should also drop support for outdated Node versions, which would be a breaking version. I don't mind doing that, upgrading should be easy enough.

Can you update
 "engines": { 
   "node": ">=6" 
 }, 

(Bump it to 8 for now)

And the .travis.yml file to test in Node 12 (which is LTS now)?
```
Hi,

Sorry for the constant updates:
This should be the final one:

updated travis.yml a bit
the tests themselves can be run with node v8

the docs rely on something with hapi and if you try to bring them up to v8 the tests fail.
```
tap needs to be left on v12.1.1

"engines": {
    "node": ">=8"
  },

```
npm test completes without errors

gr2m (approved):
Thanks a lot @dhuang612  💐

PR #894, opened 2020-01-23
Update dependencies and devDependencies

alanmatw:


gr2m:
Thanks Alan, but there is a lot of work needed besides updating the dependencies. Upgrading to hapi 18 from hapi 16 alone will not be easy. You have to make sure that the tests are all passing

alanmatw:
I wasn't aware of that, anyway thanks for the replay.

PR #897, opened 2020-02-24
docs: Add 'dependencies' field in package.json

linus345:
Add `dependencies` field in the quickstart `package.json` example to make it more accurate.


gr2m (approved):
Good catch, thank you!

PR #908, opened 2020-10-05
Update index.html

ghost:
Added Bootstrap for better view. Assigned container class for margin and padding along the borders and made the site page responsive using vw.

### DIFF
--- a/public/index.html
+++ b/public/index.html
@@ -4,10 +4,12 @@
     <meta charset="utf-8">
     <meta name="viewport" content="width=device-width, initial-scale=1.0">
     <title>Hoodie</title>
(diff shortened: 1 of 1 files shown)


public/index.html
```diff
@@ -4,10 +4,12 @@
     <meta charset="utf-8">
     <meta name="viewport" content="width=device-width, initial-scale=1.0">
     <title>Hoodie</title>
+    <link rel="stylesheet" href="https://stackpath.bootstrapcdn.com/bootstrap/4.3.1/css/bootstrap.min.css" integrity="sha384-ggOyR0iXCbMQv3Xipma34MD+dH/1fQ784/j6cY/iJTQUOhcWr7x9JvoRxT2MZw1T" crossorigin="anonymous">
```
Richard-Addicott, line 7: Is there a way that you can reference the files locally? 
By referencing an external website, you are creating a dependency for an internet connection to get the relevant styles on load.

public/index.html
```diff
@@ -4,10 +4,12 @@
...
     <link rel="stylesheet" type="text/css" href="style.css">
   </head>
   <body>
-    <h1>Well done, you made it!</h1>
+      <div class="container" style="font-size:2.5vw;">
```
Richard-Addicott, line 11: Try to not use inline style attributes on elements.

PR #929, opened 2022-06-09
Setup step 1 divided

nidhi-17:


orgrune (commented):
Makes it complicated and is not necessary. I recommend leaving it as is and closing pr.

PR #934, opened 2022-08-20
added alt's to all the images I could find

tigerpa616:
I hope this helps


likhith4:
how many img's attribute did you find to add alt attribute I could only find one

